PR #7, opened 2021-07-31
Switch to CMake

### LINKED ISSUE #3: Migrate to CMake
Using a different build system has been discussed before. There are many reasons why we should use CMake instead of pure GNU Make:
1. It's the industry standard and thus a tool worth getting used to.
2. It's more transparent. Since it takes the builds to a higher level, it's easier to modify and manage.
3. IDEs are able to load CMake projects and improve their support based on that.

We should also consider upgrading to C++20 and using Clang.

MiltFra:
Fixes #3.

# ToDO
- [x] set up basic CMake project for 64-bit Linux
- [ ] complete cross-compilation
  - [ ] 32-bit
  - [ ] ARM
  - [ ] Windows
  - [ ] MacOS
- [x] compile tests
- [ ] create easy to use commands
- [ ] update CI
- [ ] enforce clang-tidy
- [x] enforce clang-format
- [ ] document build system


MiltFra:
Closed in favour of #8.

PR #28, opened 2021-09-24
MGT: Move files from test/src to test

MiltFra:
Since the only things in the `test` folder are test source files, there is no point in having a subdirectory `test/src`. It just adds irrelevant structure.


pmorande27 (approved):

suryanshmanocha (approved):
Fine by me

PR #37, opened 2021-10-17
Attempted to reduce redundancy in exception throwing in math/kalman_m…

flori-margaritescu:
…ultivariate.cpp

## Description
Create helper functions for throwing exceptions regarding incorrect matrix dimensions to reduce redundancy in code.

## Changes
- Create two helper functions to replace throwing "invalid_argument" exceptions relating to incorrect dimensions


MiltFra:
Now the issue is that you are adding a string and an integer which is not ideal. You should be using [`std::format`](https://en.cppreference.com/w/cpp/utility/format/format) instead.

PR #44, opened 2021-10-18
Mtc rpm_regulator.[ch]pp cleanup

TomLonergan03:
## Description
Correcting rpm_regulator.[ch]pp files to align to style guide
Changes `#includes` and function calls for rpm_regulator.[ch]pp in state_processor.[ch]pp to support these changes
## Changes
- Corrected file names
- Fixed variable names
- added `consts` where necessary
- Fixed class names
- Fixed type error in `calculateRpm`


src/propulsion/rpm_regulator.cpp
```diff
@@ -14,14 +14,13 @@ int32_t RpmRegulator::calculateRpm(const data::nav_t actual_velocity, const int3
                                    const int32_t actual_current, const int32_t actual_temperature)
 {
   const int32_t optimal_rpm = calculateOptimalRpm(actual_velocity);
-  if (actual_temperature > kMaximumTemperature) { return actual_rpm; }
-  if (actual_current < kMaximumCurrent && actual_rpm < optimal_rpm) {
-    return actual_rpm + step(optimal_rpm, actual_rpm);
+  if (actual_current < kMaximumCurrent && actual_rpm == optimal_rpm
```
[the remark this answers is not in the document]
MiltFra: Bump @TomLonergan03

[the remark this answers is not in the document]
TomLonergan03: Whoops sorry I forgot about this
Should be sorted now I think

src/propulsion/rpm_regulator.cpp
```diff
@@ -0,0 +1,43 @@
...
+                                   const int32_t actual_current, const int32_t actual_temperature)
+{
+  const int32_t optimal_rpm = calculateOptimalRpm(actual_velocity);
+  if (actual_current > kMaximumCurrent || actual_rpm == optimal_rpm
+      || actual_temperature > kMaximumTemperature) {
+    return actual_rpm;
+  }
+  const int32_t target = actual_rpm + step(optimal_rpm, actual_rpm);
+  if (target < 0) { return 0; }
+  return target;
```
MiltFra, line 23: `return std::max(target, 0);`

src/propulsion/rpm_regulator.cpp
```diff
@@ -19,8 +19,7 @@ int32_t RpmRegulator::calculateRpm(const data::nav_t actual_velocity, const int3
     return actual_rpm;
   }
   const int32_t target = actual_rpm + step(optimal_rpm, actual_rpm);
-  if (target < 0) { return 0; }
-  return target;
+  return std::max(target,0);
```
MiltFra, line 22: pretty sure there's a space missing here... sorry :D

MiltFra (approved):

src/propulsion/rpm_regulator.cpp
```diff
@@ -0,0 +1,42 @@
...
+      || actual_temperature > kMaximumTemperature) {
+    return actual_rpm;
+  }
+  const int32_t target = actual_rpm + step(optimal_rpm, actual_rpm);
+  return std::max(target, 0);
+}
+
+int32_t RpmRegulator::calculateOptimalRpm(const data::nav_t actual_velocity)
+{
+  return std::round(0.32047 * actual_velocity * actual_velocity + 297.72578 * actual_velocity
```
suryanshmanocha, line 27: Sorry - I don't have much context here, but where are these number derived from? Generally 'magic numbers' are bad practise. I would store these as constant expressions instead IMO.

src/propulsion/rpm_regulator.cpp
```diff
@@ -0,0 +1,42 @@
...
+                    + 1024.30824);
+}
+
+int32_t RpmRegulator::step(const int32_t optimal_rpm, const int32_t actual_rpm)
+{
+  if (actual_rpm < optimal_rpm) { return std::round(optimal_rpm * 0.1); }
```
suryanshmanocha, line 33: Ternery might be cleaner:
`return (actual_rpm < optimal_rpm) ? return std::round(optimal_rpm * 0.1) : std::round(optimal_rpm * -0.05);`

suryanshmanocha (changes requested):
Minor nitpicks

src/propulsion/rpm_regulator.cpp
MiltFra: That's a very valid question, one that we found an answer for: through simulations, we determined the optimal frequencies for different velocities. To make this data usable, somebody fit a quadratic polynomial to those data points and this is the result. I agree though, there should at least be a comment. Please don't extract constants because this is going to change anyways - don't make it look more sophisticated than it is.

src/propulsion/rpm_regulator.cpp
MiltFra: Thanks for reminding me to ban the ternary operator in the new repo. Jokes aside, I'm not convinced this operator has ever led to anything good.

src/propulsion/rpm_regulator.cpp
suryanshmanocha: What do you mean it's going to change ?
Should there not be a //TODO: in that case?

src/propulsion/rpm_regulator.cpp
suryanshmanocha: Fair enough - there seems to be quite some debate on whether they actually improve readability

src/propulsion/rpm_regulator.cpp
MiltFra: The `// TODO` would be "figure out motor controllers" so I don't think there's any point in having it... Like, we have no clue what the final solution is going to look like so it doesn't matter too much.

src/propulsion/rpm_regulator.cpp
```diff
@@ -25,7 +25,7 @@ int32_t RpmRegulator::calculateRpm(const data::nav_t actual_velocity, const int3
 int32_t RpmRegulator::calculateOptimalRpm(const data::nav_t actual_velocity)
 {
   return std::round(0.32047 * actual_velocity * actual_velocity + 297.72578 * actual_velocity
-                    + 1024.30824);
```
MiltFra: Try putting comments on the line above so you don't contribute to the line length, possibly exceeding it. It makes the auto-formatting more difficult.

suryanshmanocha (approved):

PR #61, opened 2021-10-21
NAV: IMU Data Logger Cleanup

callumTM:
Cleans `imu_data_logger.cpp` and  `imu_data_logger.hpp`

## Changes
- Added `const`
- Removed `using`
- Fixed types
- Fixed variable names to adhere to Style Guide

### DIFF
--- a/src/navigation/imu_data_logger.cpp
+++ b/src/navigation/imu_data_logger.cpp
@@ -18,7 +18,7 @@ void ImuDataLogger::setup(const int imu_id, const int run_id)
   snprintf(buff, sizeof(buff), "test_data/run%d/imu%d/data.csv", run_id, imu_id);
   file_path_ = buff;
   outfile_->open(file_path_);
-  *outfile_ << "arx,ary,arz,acx,acy,acz,t\n";
+  *outfile_ << "arx,ary,arz,acx,acy,acz,t" << std::endl;
 }
 
 void ImuDataLogger::setupKalman(const int imu_id, const int run_id)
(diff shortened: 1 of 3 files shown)


src/navigation/imu_data_logger.cpp
```diff
@@ -12,41 +15,60 @@ ImuDataLogger::~ImuDataLogger()
...
-void ImuDataLogger::dataToFileKalman(NavigationVector &accR, NavigationVector &accC,
-                                     NavigationVector &x, uint32_t timestamp)
+void ImuDataLogger::dataToFileKalman(const data::NavigationVector &raw_acceleration,
+                                     const data::NavigationVector &calibrated_acceleration,
+                                     const data::NavigationVector &x, const uint32_t timestamp)
 {
-  *outfile_ << accR[0] << "," << accR[1] << "," << accR[2] << "," << accC[0] << "," << accC[1]
-            << "," << accC[2] << "," << x[0] << "," << x[1] << "," << x[2] << "," << timestamp
-            << "\n";
+  *outfile_ << raw_acceleration[0] << ",";
```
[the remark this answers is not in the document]
callumTM: I believe we need to keep the order, however, I think we could do 3 for loops, each with 3 iterations i.e. one for raw acceleration, another for calibrated then the last for x?

src/navigation/imu_data_logger.hpp
```diff
@@ -4,29 +4,68 @@
...
 #include <data/data_point.hpp>
 
 namespace hyped {
-using data::DataPoint;
-using data::nav_t;
-using data::NavigationVector;
 
 namespace navigation {
 
 class ImuDataLogger {
```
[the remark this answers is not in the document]
callumTM: Is this required? I've not seen it in any of the other header files.

src/navigation/imu_data_logger.cpp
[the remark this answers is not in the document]
suryanshmanocha: That's what I meant, I think that would be better!

src/navigation/imu_data_logger.hpp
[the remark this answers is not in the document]
suryanshmanocha: I meant a brief docstring for the constructor - it should be at line 22, my bad.

src/navigation/imu_data_logger.cpp
```diff
@@ -12,41 +15,60 @@ ImuDataLogger::~ImuDataLogger()
...
+  *outfile_ << raw_acceleration[1] << ",";
```
MiltFra, line 63: Just noticed that this `ofstream` is still behind a pointer even though there's no reason for that.

callumTM: I'm not 100% sure what you mean. Could you clarify please?

MiltFra: It's declared as `std::ofstream *outfile_` but I think making it `std::ofstream outfile_` would be fine. There's no real reason to make this a pointer.

MiltFra: Then you can remove the arguments from the constructor altogether and call `outfile_.open(..)` instead of `outfile_->open(..)`.

src/navigation/imu_data_logger.cpp
```diff
@@ -1,52 +1,74 @@
...
+                                     const data::NavigationVector &calibrated_acceleration,
+                                     const data::NavigationVector &x, const uint32_t timestamp)
 {
-  *outfile_ << accR[0] << "," << accR[1] << "," << accR[2] << "," << accC[0] << "," << accC[1]
-            << "," << accC[2] << "," << x[0] << "," << x[1] << "," << x[2] << "," << timestamp
-            << "\n";
+  for (std::size_t i = 0; i < 3; i++) {
+    outfile_ << raw_acceleration[i] << ",";
+  }
+  for (std::size_t i = 0; i < 3; i++) {
```
suryanshmanocha, line 65: Sorry for going back and forth on this, maybe you can make an auxiliary function to help add commas to avoid code repetition?

So it'll look something like this:
`void ImuDataLogger::appendToFile(const data::NavigationVector &vector) {
  for (std::size_t i = 0; i < 3; i++) {
    outfile_ << vector[i] << ",";
}`

Then we just call:
 `appendToFile(raw_acceleration)
appendToFile(calibrated_acceleration)
appendToFile(x)
`

suryanshmanocha (changes requested):
Small thing only

src/navigation/imu_data_logger.cpp
MiltFra: I'm not sure I agree on this to be honest. You're right, there's repetition here, but I think it's very clear what's happening. This is not a massive function in need of refactoring so I'm tempted to leave it like that.

suryanshmanocha (approved):

MiltFra (approved):

PR #76, opened 2021-10-28
GPIO Cleanup

kshxtij:
## Description
Clean up of gpio_manager.cpp, gpio_manager.hpp, gpio_counter.cpp, gpio_counter.hpp.

Removed all references to using.

### DIFF
--- a/src/sensors/gpio_counter.hpp
+++ b/src/sensors/gpio_counter.hpp
@@ -16,7 +16,7 @@ using utils::concurrent::Thread;
 
 namespace sensors {
 
-class GpioCounter : public GpioInterface, public Thread {  // interface.hpp
+class GpioCounter : public GpioInterface, public utils::concurrent::Thread {  // interface.hpp
  public:
   GpioCounter(utils::Logger &log, int pin);
   ~GpioCounter() {}
@@ -26,7 +26,7 @@ class GpioCounter : public GpioInterface, public Thread {  // interface.hpp
 
  private:
   int pin_;
-  System &sys_;
+  utils::System &sys_;
   utils::Logger &log_;
   data::StripeCounter stripe_counter_;
 };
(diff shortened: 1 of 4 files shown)


src/sensors/gpio_manager.cpp
```diff
@@ -6,19 +6,19 @@
...
     : Thread(log),
       sys_(utils::System::getSystem()),
-      data_(Data::getInstance())
+      data_(data::Data::getInstance())
 {
   // clear HPSSRs if default is high
-  for (int i = 0; i < data::Batteries::kNumHPBatteries; i++) {
-    hp_ssr_.push_back(new GPIO(sys_.config->sensors.hp_shutoff[i], utils::io::gpio::kOut));
+  for (std::size_t i; i < data::Batteries::kNumHPBatteries; ++i) {
+    hp_ssr_.emplace_back(new utils::io::GPIO(sys_.config->sensors.hp_shutoff[i], utils::io::gpio::kOut));
```
MiltFra, line 16: this should be `hp_ssr_.emplace_back(sys.config->sensors.hp_shutoff[i], utils::io::gpio::kOut)`. Note that `new` performs a heap allocation and returns a pointer. However, we want to construct the object in the memory that has been allocated for the vector already.

MiltFra:
Huh, seems like this still requires the copy constructor, didn't know that...

src/sensors/gpio_counter.hpp
```diff
@@ -16,7 +16,7 @@ using utils::concurrent::Thread;
 
 namespace sensors {
 
-class GpioCounter : public GpioInterface, public Thread {  // interface.hpp
+class GpioCounter : public GpioInterface, public utils::concurrent::Thread {  // interface.hpp
  public:
   GpioCounter(utils::Logger &log, int pin);
```
suryanshmanocha, line 21: const int

suryanshmanocha (changes requested):

src/sensors/gpio_counter.hpp
MiltFra: Not even sure `int` is the right type here...

MiltFra:
Closed in favour of #98.

PR #78, opened 2021-11-01
QA cleanup utils

weijuinlee:
## Changes
- Removed std::chrono::high_resolution_clock::now() from logger
- Removed getTimeMillis() as it was replaced by getTimeMicros() from timer


MiltFra (approved):
There are more things wrong but I don't want to get caught up in these files.

flori-margaritescu (approved):

PR #79, opened 2021-11-01
QA debug

weijuinlee:
Added Socket and Client Socket


MiltFra:
Seems like you added a few more changes than you should have...

src/debug/server.py
```diff
@@ -0,0 +1,43 @@
...
+    while connected:
+        msg_length = conn.recv(HEADER).decode(FORMAT)
+        if msg_length:
+            msg_length = len(msg_length)
+            msg = conn.recv(msg_length).decode(FORMAT)
+            if msg == DISCONNECT_MESSAGE:
+                connected = False
+
+            print(f"[{addr}] {msg}")
+            conn.send("Msg received".encode(FORMAT))
```
MiltFra, line 27: Don't abreviate too much. It's Message not Msg, it's connection not conn, ...

src/debug/client.cpp
```diff
@@ -0,0 +1,39 @@
+#include <stdio.h>
+#include <sys/socket.h>
+#include <arpa/inet.h>
+#include <unistd.h>
+#include <string.h>
+#define PORT 5050
```
MiltFra, line 6: `static constexpr uint16_t kPort = 5050`

src/debug/client.cpp
```diff
@@ -0,0 +1,39 @@
...
+   
+int main(int argc, char const *argv[])
+{
+    int sock = 0, valread;
+    struct sockaddr_in serv_addr;
+    char *hello = "Client";
```
MiltFra, line 12: what's this?!

src/debug/client.cpp
```diff
@@ -0,0 +1,39 @@
...
+#include <sys/socket.h>
+#include <arpa/inet.h>
+#include <unistd.h>
+#include <string.h>
+#define PORT 5050
+   
+int main(int argc, char const *argv[])
+{
+    int sock = 0, valread;
+    struct sockaddr_in serv_addr;
```
MiltFra, line 11: drop the `struct` here

src/debug/client.cpp
```diff
@@ -0,0 +1,39 @@
+#include <stdio.h>
+#include <sys/socket.h>
+#include <arpa/inet.h>
+#include <unistd.h>
+#include <string.h>
+#define PORT 5050
+   
+int main(int argc, char const *argv[])
+{
+    int sock = 0, valread;
```
MiltFra, line 10: Don't mix declarations and assignments.

MiltFra (changes requested):
I've pointed out a couple of things, but there are more things wrong. Most prominently, I really dislike the file structure. (a) you put the client.cpp and the server.py in the same folder of a C++ (!!!) project. The Python part is going to be running on a different machine. I'd either put it into the base station repo or make a new one (preference is the latter because the base station repo already is a bit messy). Set up a proper python package that runs the server, use type annotation and follow best practices in general. As for the client.cpp, there's a main function in that file. Why? You're main is going to be in run/debug.cpp where you will create a TCP client. By the way, telemetry already has an implementation for this so I don't see a reason why you can't use that. Don't reinvent the wheel.

PR #80, opened 2021-11-01
Imu comments

AlbinGrip99:
## Description
Give a brief summary of what this PR contributes 

 *Click up link* [](<link here>)

## Changes
- change 1
- change 2
- change 3

## Additional Info 
screenshots, commands to run etc.

### DIFF
--- a/src/sensors/gpio_counter.cpp
+++ b/src/sensors/gpio_counter.cpp
@@ -2,6 +2,7 @@
 
 #include <stdio.h>
 
+//Where to find the utils??
 #include <utils/io/gpio.hpp>
 #include <utils/timer.hpp>
 
@@ -39,6 +40,7 @@ void GpioCounter::run()
   }
 }
 
+//Is this the method where the count is updated?? What does the operational part do?
 void GpioCounter::getData(StripeCounter *stripe_counter)
 {
   stripe_counter->count       = stripe_counter_.count;
--- a/src/sensors/imu.cpp
+++ b/src/sensors/imu.cpp
@@ -144,7 +144,7 @@ void Imu::enableFifo()
   kFrameSize_ = 6;
 }
 
-bool Imu::whoAmI()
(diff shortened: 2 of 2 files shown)


src/sensors/imu.cpp
```diff
@@ -179,14 +179,14 @@ void Imu::selectBank(uint8_t switch_bank)
   log_.DBG1("Imu", "User bank switched to %u", user_bank_);
 }
 
-void Imu::writeByte(uint8_t write_reg, uint8_t write_data)
+void Imu::writeByte(uint8_t write_reg, uint8_t write_data) //MISO data shift
 {
   // ',' instead of ';' is to inform the compiler not to reorder function calls
```
MiltFra, line 184: https://en.cppreference.com/w/cpp/atomic/memory_order

src/sensors/imu.cpp
```diff
@@ -179,14 +179,14 @@ void Imu::selectBank(uint8_t switch_bank)
   log_.DBG1("Imu", "User bank switched to %u", user_bank_);
 }
 
-void Imu::writeByte(uint8_t write_reg, uint8_t write_data)
+void Imu::writeByte(uint8_t write_reg, uint8_t write_data) //MISO data shift
```
MiltFra, line 182: Add this sort of information in the header, or in the function body. Putting it after the function signature may screw formatting and looks inconsistent.

src/sensors/imu.cpp
MiltFra: https://www.cplusplus.com/reference/atomic/atomic_signal_fence/

src/sensors/gpio_counter.cpp
```diff
@@ -2,6 +2,7 @@
 
 #include <stdio.h>
 
+//Where to find the utils??
```
MiltFra, line 5: In src/utils? Please don't leave comments like this in the code base :D

MiltFra (changes requested):

PR #82, opened 2021-11-02
NAV: main_log.cpp/hpp Cleanup

callumTM:
Cleans `main_log.cpp `and `main_log.hpp`

## Changes
- changed `int` to `std::size_t` for loops
- removed `using`
- renamed `acc` to `acceleration`
- changed [i] to .at(i)

### DIFF
--- a/src/navigation/main_log.cpp
+++ b/src/navigation/main_log.cpp
@@ -12,7 +12,7 @@ MainLog::MainLog(const uint8_t id, Logger &log)
     : Thread(id, log),
       log_(log),
       sys_(System::getSystem()),
-      data_(Data::getInstance())
+      data_(data::Data::getInstance())
 {
   log_.INFO("NAV", "Logging initialising");
   calibrateGravity();
@@ -54,11 +54,14 @@ void MainLog::run()
     DataPoint<ImuDataArray> sensor_readings = data_.getSensorsImuData();
     for (std::size_t i = 0; i < data::Sensors::kNumImus; ++i) {
       // Apply calibrated correction
-      NavigationVector acc     = sensor_readings.value[i].acc;
-      NavigationVector acc_cor = acc - gravity_calibration_[i];
-      log_.DBG("NAV", "%.3f %.3f %.3f / %.3f %.3f %.3f", acc[0], acc[1], acc[2], acc_cor[0],
-               acc_cor[1], acc_cor[2]);
-      imu_loggers_[i].dataToFile(acc_cor, acc, sensor_readings.timestamp);
+      NavigationVector acceleration
+        = sensor_readings.value[i].acc;  // TODO: .acc should be renamed to acceleration, but must
+                                         // be changed in data.hpp and other areas as
+      NavigationVector calibrated_acceleration = acceleration - gravity_calibration_[i];
+      log_.DBG("NAV", "%.3f %.3f %.3f / %.3f %.3f %.3f", acceleration[0], acceleration[1],
+               acceleration[2], calibrated_acceleration[0], calibrated_acceleration[1],
+               calibrated_acceleration[2]);
+      imu_loggers_[i].dataToFile(calibrated_acceleration, acceleration, sensor_readings.timestamp);
     }
   }
 }
--- a/src/navigation/main_log.cpp
+++ b/src/navigation/main_log.cpp
@@ -50,7 +50,6 @@ void MainLog::run()
   log_.INFO("NAV", "Logging starting");
 
   while (sys_.running_) {
-    DataPoint<ImuDataArray> sensor_readings = data_.getSensorsImuData();
     data::DataPoint<ImuDataArray> sensor_readings = data_.getSensorsImuData();
     for (std::size_t i = 0; i < data::Sensors::kNumImus; ++i) {
       // Apply calibrated correction
--- a/src/navigation/main_log.cpp
+++ b/src/navigation/main_log.cpp
@@ -8,7 +8,7 @@ typedef std::array<ImuData, data::Sensors::kNumImus> ImuDataArray;
 
 namespace navigation {
 
-MainLog::MainLog(uint8_t id, Logger &log)
+MainLog::MainLog(const uint8_t id, Logger &log)
     : Thread(id, log),
       log_(log),
       sys_(System::getSystem()),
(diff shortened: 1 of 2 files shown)


src/navigation/main_log.cpp
```diff
@@ -31,14 +31,14 @@ void MainLog::calibrateGravity()
...
+    for (std::size_t j = 0; j < data::Sensors::kNumImus; ++j) {
       online_array[j].update(sensor_readings.value[j].acc);
     }
     Thread::sleep(1);
   }
-  for (int j = 0; j < data::Sensors::kNumImus; ++j) {
+  for (std::size_t j = 0; j < data::Sensors::kNumImus; ++j) {
     gravity_calibration_[j] = online_array[j].getMean();
     log_.INFO("NAV",
               "Update: g=(%.5f, %.5f, %.5f)",  // NOLINT
```
MiltFra, line 44: Remove the `// NOLINT`

src/navigation/main_log.cpp
```diff
@@ -31,14 +31,14 @@ void MainLog::calibrateGravity()
...
     DataPoint<ImuDataArray> sensor_readings = data_.getSensorsImuData();
-    for (int j = 0; j < data::Sensors::kNumImus; ++j) {
+    for (std::size_t j = 0; j < data::Sensors::kNumImus; ++j) {
       online_array[j].update(sensor_readings.value[j].acc);
     }
     Thread::sleep(1);
   }
-  for (int j = 0; j < data::Sensors::kNumImus; ++j) {
+  for (std::size_t j = 0; j < data::Sensors::kNumImus; ++j) {
     gravity_calibration_[j] = online_array[j].getMean();
```
MiltFra, line 42: Use `.at(i)` instead of `[i]`

src/navigation/main_log.hpp
```diff
@@ -39,7 +44,7 @@ class MainLog : public Thread {
 
   Logger &log_;
```
MiltFra, line 45: `utils::`

src/navigation/main_log.hpp
```diff
@@ -29,8 +28,14 @@ using utils::math::OnlineStatistics;
...
-class MainLog : public Thread {
+class MainLog : public utils::concurrent::Thread {
  public:
+  /**
+   * @brief Construct a new Main Log object
+   *
+   * @param id log ID
+   * @param log Logger object
+   */
   explicit MainLog(uint8_t id, Logger &log);
```
MiltFra, line 39: This doens't agree with the cpp file.

src/navigation/main_log.hpp
```diff
@@ -18,7 +18,6 @@
 
 namespace hyped {
 
-using data::Data;
 using data::ImuData;
```
MiltFra, line 21: lots of `using`

MiltFra (changes requested):

src/navigation/main_log.cpp
```diff
@@ -28,20 +28,20 @@ MainLog::MainLog(uint8_t id, Logger &log)
...
-    Thread::sleep(1);
+    utils::concurrent::Thread::sleep(1);
   }
   for (std::size_t j = 0; j < data::Sensors::kNumImus; ++j) {
-    gravity_calibration_[j] = online_array[j].getMean();
-    log_.INFO("NAV",
-              "Update: g=(%.5f, %.5f, %.5f)",  // NOLINT
-              gravity_calibration_[j][0], gravity_calibration_[j][1], gravity_calibration_[j][2]);
+    gravity_calibration_.at(j) = online_array.at(j).getMean();
+    log_.INFO("NAV", "Update: g=(%.5f, %.5f, %.5f)", gravity_calibration_.at(j)[0],
```
callumTM, line 43: This can either be `gravity_calibration_.at(j)[0]` or what it was originally `gravity_calibration_[j][0]`. The first is clearer but the second is closer to the preferred style.

src/navigation/main_log.hpp
callumTM: Should the log be described just as "System logger"?

src/navigation/main_log.cpp
MiltFra: I prefer the former. In any case, I think we should provide functionality to log vectors more consistently.

src/navigation/main_log.hpp
```diff
@@ -18,31 +18,28 @@
...
+class MainLog : public utils::concurrent::Thread {
  public:
-  explicit MainLog(uint8_t id, Logger &log);
+  /**
+   * @brief Construct a new Main Log object
+   *
+   * @param id log ID
+   * @param log Logger object
+   */
+  explicit MainLog(uint8_t id, utils::Logger &log);
```
MiltFra, line 31: What I mean by disagreement is that this should be `explicit MainLog(const uint8_t id, utils::Logger &log)`; exactly as it is in the `.cpp` file.

src/navigation/main_log.cpp
```diff
@@ -4,21 +4,21 @@
 
 namespace hyped {
 
-typedef std::array<ImuData, data::Sensors::kNumImus> ImuDataArray;
+typedef std::array<data::ImuData, data::Sensors::kNumImus> ImuDataArray;
```
MiltFra, line 7: I'm not a fan of this `typedef` but for now you should convert it to a `using`:

```cpp
using ImuDataArray = std::array<data::ImuData, data::Sensors::kNumImus>;
```

This is considered better practice because it's more readable and has a few more features.

src/navigation/main_log.cpp
```diff
@@ -53,11 +53,15 @@ void MainLog::run()
     data::DataPoint<ImuDataArray> sensor_readings = data_.getSensorsImuData();
     for (std::size_t i = 0; i < data::Sensors::kNumImus; ++i) {
       // Apply calibrated correction
-      NavigationVector acc     = sensor_readings.value[i].acc;
-      NavigationVector acc_cor = acc - gravity_calibration_[i];
-      log_.DBG("NAV", "%.3f %.3f %.3f / %.3f %.3f %.3f", acc[0], acc[1], acc[2], acc_cor[0],
-               acc_cor[1], acc_cor[2]);
-      imu_loggers_[i].dataToFile(acc_cor, acc, sensor_readings.timestamp);
+      data::NavigationVector acceleration
+        = sensor_readings.value.at(i).acc;  // TODO: .acc should be renamed to acceleration, but
```
MiltFra, line 57: This TODO should go on the line above because it's messing up the formatting.

MiltFra (approved):

suryanshmanocha (approved):

weijuinlee (commented):
Looks alright to me and I agree with the use of scope resolution to avoid conflict.

PR #84, opened 2021-11-04
started implementing additional and status data packing

adenhaus:
## Description
Started implementing `packAdditionalData` and `packStatusData` in `writer.cpp`

## Additional Info 
Goal is to pack entire (relevant) central data structure from `data.hpp`. Maybe will combine `crucial`, `additional` and `status` data in the future but for now just getting it packed.

`packStatusData` says it's not editable, but I edited anyway. Only reason it's marked as not editable is because of the GUI, which expects a certain input from the status data I believe. But that can be fixed once we're actually sending all the data we want.

## TODO
- [x] Add tests for JSON serialising and deserialising


src/telemetry/writer.cpp
```diff
@@ -25,9 +25,98 @@ void Writer::packAdditionalData()
...
+  // Sensors (emergency brakes and batteries included)
+  data::Sensors sensors_data        = data_.getSensorsData();
+  data::Batteries batteries_data    = data_.getBatteriesData();
+  data::EmergencyBrakes brakes_data = data_.getEmergencyBrakesData();
+  startList("Sensors");
+  add("brakes_retracted", brakes_data.brakes_retracted);
+  add("temperature", data_.getTemperature());
+
+  startList("Low Power Batteries");
+  for (int i = 0; i < batteries_data.kNumLPBatteries; i++) {
```
MiltFra, line 56: `++i`

src/telemetry/writer.cpp
```diff
@@ -25,9 +25,98 @@ void Writer::packAdditionalData()
...
+    add("lp_battery_" + battery_num + "_high_volt_cell",
+        batteries_data.low_power_batteries[i].high_voltage_cell);
+    add("lp_battery_" + battery_num + "_imd_fault",
+        batteries_data.low_power_batteries[i].imd_fault);
+    endList();
+  }
+  endList();
+
+  startList("High Power Batteries");
+  for (int i = 0; i < batteries_data.kNumHPBatteries; i++) {
```
MiltFra, line 79: `std::size_t` instead of `int`

src/telemetry/writer.cpp
```diff
@@ -56,7 +145,19 @@ void Writer::packStatusData()
...
+  // Module statuses
+  data::Navigation nav_data = data_.getNavigationData();
+  add("nav_status", nav_data.module_status);
+  data::Motors motors_data = data_.getMotorData();
+  add("motors_status", motors_data.module_status);
+  data::Sensors sensors_data = data_.getSensorsData();
+  add("sensors_status", sensors_data.module_status);
+  data::Telemetry telemetry_data = data_.getTelemetryData();
+  add("telemetry_status", telemetry_data.module_status);
+  data::EmergencyBrakes brakes_data = data_.getEmergencyBrakesData();
```
MiltFra, line 157: I'd make this `const auto` instead of `data::EmergencyBrakes`

src/telemetry/writer.cpp
```diff
@@ -25,9 +25,98 @@ void Writer::packAdditionalData()
...
+  add("brakes_retracted", brakes_data.brakes_retracted);
+  add("temperature", data_.getTemperature());
+
+  startList("Low Power Batteries");
+  for (int i = 0; i < batteries_data.kNumLPBatteries; i++) {
+    std::string battery_num = std::to_string(i);
+    startList("Low Power Battery " + battery_num);
+    add("lp_battery_" + battery_num + "_ave_temp",
+        batteries_data.low_power_batteries[i].average_temperature);
+    add("lp_battery_" + battery_num + "_volt", batteries_data.low_power_batteries[i].voltage);
```
MiltFra, line 61: can we not have a JSON array instead of putting the index in the key? Just structure it the same way as we do it in `data`.

MiltFra (changes requested):

src/telemetry/writer.cpp
adenhaus: Not sure what you mean. In `data` it's just a struct? There's no `add` overload in `writer` that supports arrays right now, so I think every variable has to be added individually.

MiltFra: I mean, the writer is part of our code so you can just add that functionality?

src/telemetry/writer.cpp
MiltFra: all

src/telemetry/writer.cpp
adenhaus: `batteries_data.high_power_batteries` and `batteries_data.low_power_batteries` are arrays of structs. RapidJSON does not support serialising structs, so it isn't an array issue, the struct has to be manually packed, and having the index tells us which battery it is anyway. https://github.com/Tencent/rapidjson/issues/1299

MiltFra (changes requested):
Doesn't build because of GTest.

MiltFra:
Have you tried running this? I doubt this works.

adenhaus:
> Have you tried running this? I doubt this works. You'll have to call `json_writer_.StartObject` somewhere. Right now you're just adding the contents of the battery struct to the array consecutively, instead of adding separate JSON objects for each struct.

I'm pretty sure it would work because this is how the rest of the data points are being added. `StartObject` is called by `start` in `sendloop.cpp`, then all the `pack` functions are called, and inside them `add` is used to add stuff.

Anyway, I've rewritten it so that there is an `add` overload specifically for the `BatteryData` struct, so each one is its own `json` object. This approach doesn't really seem aligned to what the rest of the code is doing, because other structs don't really have their own `add` overloads, `add` is just used to add every one of their individual fields, which is what I was doing for `BatteryData` before. On the other hand, even though it's not how the others do it, this might be the best approach for `BatteryData` specifically because it's the only case where we have structs in a list.

Let me know what you think.

src/telemetry/writer.cpp
```diff
@@ -25,9 +25,64 @@ void Writer::packAdditionalData()
...
+  // Navigation
+  const auto nav_data = data_.getNavigationData();
+  startList("Navigation");
+  add("braking_distance", 0.0, 1250.0, "m", nav_data.braking_distance);
+  add("displacement", 0.0, 1250.0, "m", nav_data.displacement);
+  add("emergency_braking_distance", 0.0, 1250.0, "m", nav_data.emergency_braking_distance);
+  endList();
+
+  // Sensors (emergency brakes and batteries included)
+  const auto sensors_data   = data_.getSensorsData();
```
MiltFra, line 48: I don't see this being used. In fact, I like the structure above where you get one part of the CDS and pack it and then the next and so on. I'd stick to that.

src/telemetry/writer.cpp
```diff
@@ -56,7 +111,19 @@ void Writer::packStatusData()
   json_writer_.Key("status_data");
   json_writer_.StartArray();
 
-  // TODO(everyone): add all required data points
+  // Module statuses
+  const auto nav_data = data_.getNavigationData();
```
MiltFra, line 115: Try and be a bit more conservative with obtaining data. It's a relatively expensive operation and here you're just taking the module status. You may want to consider making functions

```cpp
void Writer::packNavigationData() {}
void Writer::packMotorsData() {}
/* ... */
```
You can then even pack each part into its own object, the way it's being done in the CDS.

MiltFra: Each of those functions would then require exactly one call to `get<Module>Data` and extract all the information at once. By the way, this also ensures that the data within a certain section does not get out of sync due to multiple calls (in between which changes may occur).

MiltFra (changes requested):

src/telemetry/writer.cpp
```diff
@@ -25,9 +25,64 @@ void Writer::packAdditionalData()
...
+  endList();
 
-  // Edit above
+  // Navigation
+  const auto nav_data = data_.getNavigationData();
+  startList("Navigation");
+  add("braking_distance", 0.0, 1250.0, "m", nav_data.braking_distance);
+  add("displacement", 0.0, 1250.0, "m", nav_data.displacement);
+  add("emergency_braking_distance", 0.0, 1250.0, "m", nav_data.emergency_braking_distance);
+  endList();
```
adenhaus, line 45: @MiltFra this `add` overload packs the minimum and maximum values, and the unit. Not sure what the purpose of this is... but it definitely seems unnecessary to send it every time, as they're constants. Some in `crucialData` are in the same format, I assume the min and max values are helpful for visualising the data. Would it make sense to send these values just once at the beginning instead?

MiltFra: Yes, I agree with that.

weijuinlee (approved):

src/telemetry/sendloop.cpp
```diff
@@ -27,10 +27,13 @@ void SendLoop::run()
...
-    writer.packStatusData();
-    writer.packAdditionalData();
+    writer.packTelemetryData();
+    writer.packSensorsData();
+    writer.packMotorData();
+    writer.packStateMachineData();
+    writer.packNavigationData();
     writer.end();
+
     data::Telemetry telemetry_data = data_.getTelemetryData();
```
MiltFra, line 37: I'd move this inside the if because we don't need it outside.

src/telemetry/writer.cpp
```diff
@@ -21,130 +25,156 @@ void Writer::packTime()
...
-  json_writer_.String(convertStateMachineState(value).c_str());
+  json_writer_.Key("critical_failure");
+  json_writer_.Bool(sm_data.critical_failure);
+  json_writer_.Key("current_state");
+  json_writer_.String(convertStateMachineState(sm_data.current_state).c_str());  // Crucial
   json_writer_.EndObject();
 }
 
-void Writer::add(const std::string name, data::ModuleStatus value)
+void Writer::packBattery(std::string name, data::BatteryData battery)
```
MiltFra, line 152: `const` and references please

src/telemetry/writer.hpp
```diff
@@ -29,12 +29,16 @@ class Writer {
  public:
   explicit Writer(data::Data &data);
 
-  // separate functions that allow to better manage data points based on their purpose
+  // functions to pack timestamp and number of package
   void packTime();
   void packId(uint16_t id);
```
MiltFra, line 34: `const`

src/telemetry/writer.hpp
```diff
@@ -51,13 +55,8 @@ class Writer {
...
-  void add(const std::string name, float min, float max, const std::string unit, float value);
-  void add(const std::string name, bool value);
-  void add(const std::string name, const std::string value);
-  void add(const std::string name, data::State value);
-  void add(const std::string name, data::ModuleStatus value);
+  // functions to pack internal CDS structs/types
+  void packBattery(std::string name, const data::BatteryData battery);
 
   // starts and ends lists, which allow to structure the data
   void startList(const std::string name);
```
MiltFra, line 62: reference

MiltFra (changes requested):
Also, why is `num_packages_sent` of type `uint16_t`? I think we can afford to make it 32 bits; better safe than sorry.

src/telemetry/writer.cpp
```diff
@@ -21,130 +25,156 @@ void Writer::packTime()
...
+  json_writer_.Key(name.c_str());
   json_writer_.StartObject();
-  json_writer_.Key("name");
-  json_writer_.String(name.c_str());
-  json_writer_.Key("value");
-  json_writer_.String(convertModuleStatus(value).c_str());
+  json_writer_.Key("average_temp");
+  json_writer_.Int(battery.average_temperature);
+  json_writer_.Key("voltage");
+  json_writer_.Int(battery.voltage);
```
SnickeyX, line 159: If this is intentional then it's perfectly fine, but both voltage and current are logged twice here?

src/telemetry/writer.cpp
```diff
@@ -21,130 +25,156 @@ void Writer::packTime()
...
+  json_writer_.Key("voltage");
+  json_writer_.Int(battery.voltage);
+  json_writer_.Key("current");
+  json_writer_.Int(battery.current);
+  json_writer_.Key("current");
```
SnickeyX, line 164: If this is intentional this it's perfectly fine, but voltage and current are logged twice?

adenhaus: Cheers that's what happens when you copy paste code, kids

weijuinlee (approved):

src/telemetry/sender.cpp
```diff
@@ -19,19 +19,23 @@ void Sender::run()
...
-    writer.packId(num_packages_sent);
-    writer.packCrucialData();
-    writer.packStatusData();
-    writer.packAdditionalData();
-    writer.end();
+
     data::Telemetry telemetry_data = data_.getTelemetryData();
     if (!client_.sendData(writer.getString())) {
+      writer.start();
+      writer.packTime();
```
MiltFra, line 30: Why is this inside the `if`?

src/telemetry/writer.cpp
```diff
@@ -20,130 +24,152 @@ void Writer::packTime()
...
-  json_writer_.String(convertStateMachineState(value).c_str());
+  json_writer_.Key("critical_failure");
+  json_writer_.Bool(sm_data.critical_failure);
+  json_writer_.Key("current_state");
+  json_writer_.String(convertStateMachineState(sm_data.current_state).c_str());  // Crucial
   json_writer_.EndObject();
 }
 
-void Writer::add(const std::string name, data::ModuleStatus value)
+void Writer::packBattery(std::string name, const data::BatteryData &battery)
```
MiltFra, line 151: `const std::string &name`? Try if that works with your current code, otherwise don't change this.

MiltFra (changes requested):

src/telemetry/sender.cpp
adenhaus: Think I misunderstood one of your previous comments. Did you want `data::Telemetry telemetry_data = data_.getTelemetryData();` inside the `if`?

MiltFra: What we want is `data_.getTelemetryData()` before the packing, but outside the `if`. As far as I can tell the `if` only checks for errors.

src/telemetry/sendloop.cpp
adenhaus: @MiltFra see here

src/telemetry/sender.cpp
adenhaus: Yeah but `getTelemetryData()` is only called so we can set `ModuleStatus` to `kCriticalFailure` if there is an error, so it can probably go inside the `if` which is what I think you were asking for.

adenhaus:
I've written a test file but the headers have red squiggles under, `cannot open source file "telemetry/writer.hpp"` etc. Tried building it but still same issue. What do I need to do to make it find the headers?

adenhaus:
> I've written a test file but the headers have red squiggles under, `cannot open source file "telemetry/writer.hpp"` etc. Tried building it but still same issue. What do I need to do to make it find the headers?

@MiltFra

src/telemetry/sender.cpp
MiltFra: Oh, you're absolutely right. You should move `getTelemetryData` inside and leave the packing outside if I'm not mistaken.

MiltFra:
> > I've written a test file but the headers have red squiggles under, `cannot open source file "telemetry/writer.hpp"` etc. Tried building it but still same issue. What do I need to do to make it find the headers?
> 
> @MiltFra

Does it compile? try `<telemetry/writer.hpp>` instead.

adenhaus:
> > > I've written a test file but the headers have red squiggles under, `cannot open source file "telemetry/writer.hpp"` etc. Tried building it but still same issue. What do I need to do to make it find the headers?
> > 
> > 
> > @MiltFra
> 
> Does it compile? try `<telemetry/writer.hpp>` instead.

Yeah I tried `<>` instead of `""`, no difference. Yes it complies. Maybe it's just Intellisense? I'll push so you can check

test/telemetry/writer_pack_data_test.cpp
```diff
@@ -0,0 +1,97 @@
...
+  Writer writer(data_);
+  writer.start();
+  writer.packTelemetryData();
+  writer.end();
+  std::string actualJson = writer.getString();
+  std::string expectedJson = "INSERT JSON HERE"; // TODO
+  ASSERT_EQ(actualJson, expectedJson) << telemetry_packing_error;
+}
+
+TEST_F(WriterPackData, packsNavigationData)
```
MiltFra, line 39: I'd just stick to `WriterTest` to be in line with other test files.

adenhaus: That's not how it is in the other telemetry tests, shall I change them all?

adenhaus:
@MiltFra I need to pass `data` to `Writer` but not sure how to get it in the file. In other files like `sender.cpp` it's being done in the constructor. How do I go about that? (`Writer` takes `data::Data` in its constructor)

MiltFra:
> @MiltFra I need to pass `data` to `Writer` but not sure how to get it in the file. In other files like `sender.cpp` it's being done in the constructor. How do I go about that? (`Writer` takes `data::Data` in its constructor)

I don't think there's any point in doing that. Just remove it from the constructor and initialise the field with `data::Data::getInstance()`.

test/telemetry/writer_pack_data_test.cpp
adenhaus: @MiltFra

adenhaus:
> > @MiltFra I need to pass `data` to `Writer` but not sure how to get it in the file. In other files like `sender.cpp` it's being done in the constructor. How do I go about that? (`Writer` takes `data::Data` in its constructor)
> 
> I don't think there's any point in doing that. Just remove it from the constructor and initialise the field with `data::Data::getInstance()`.

Not sure what you mean. Check if what I've done now is okay: `const auto data_ = &hyped::data::Data::getInstance();` and than passing `*data_` to `Writer`. Or should I just pass `data::Data::getInstance()` to `Writer` every time?

test/telemetry/writer_pack_data_test.cpp
MiltFra: Oh, if that's the case then leave it. Also, message me on Slack because I may not see these in time.

test/telemetry/writer_pack_data.test.cpp
```diff
@@ -0,0 +1,199 @@
...
+
+TEST_F(WriterPackData, packsId)
+{
+  Writer writer(data_);
+  int id = 5;
+  writer.start();
+  writer.packId(id);
+  writer.end();
+  const std::string actualJson   = writer.getString();
+  const std::string expectedJson = "{\"id\":" + std::to_string(id) + "}";
```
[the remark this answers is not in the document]
MiltFra: Also, check this out: https://www.geeksforgeeks.org/raw-string-literal-c/

MiltFra (changes requested):

test/telemetry/writer_pack_data.test.cpp
```diff
@@ -0,0 +1,199 @@
...
+  ASSERT_EQ(actualJson, expectedJson) << "Navigation json does not match expected output.";
+  ASSERT_TRUE(writer.isValidJson()) << "Navigation json invalid.";
+}
+
+TEST_F(WriterPackData, packsSensorsData)
+{
+  setSensorsData();
+  Writer writer(data_);
+  writer.start();
+  writer.packSensorsData();
```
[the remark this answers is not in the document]
adenhaus: Yeah they're only used once, just seemed neater this way.

test/telemetry/writer_pack_data.test.cpp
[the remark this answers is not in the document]
adenhaus: I think using escape slashes and not a literal is actually useful here because in the tests, outputs are printed with the escapes. So comparing is a bit easier if they're the same. But I will change it if you think it's better.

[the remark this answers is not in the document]
MiltFra: Alright, let's stick with this then.

test/telemetry/writer_pack_data.test.cpp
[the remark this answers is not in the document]
MiltFra: Don't think it is though.

test/telemetry/writer_pack_data.test.cpp
```diff
@@ -0,0 +1,178 @@
...
+
+Data &data_ = Data::getInstance();
+
+TEST_F(WriterPackData, packsId)
+{
+  Writer writer;
+  writer.start();
+  writer.packId(5);
+  writer.end();
+  const std::string actualJson   = writer.getString();
```
MiltFra, line 25: should be `actual_json`

test/telemetry/writer_pack_data.test.cpp
```diff
@@ -0,0 +1,178 @@
+#include "data/data.hpp"
+#include "gtest/gtest.h"
+#include "telemetry/writer.hpp"
+
+using namespace hyped::telemetry;
+using namespace hyped::data;
```
MiltFra, line 6: Just seeing this now...

test/telemetry/writer_pack_data.test.cpp
```diff
@@ -0,0 +1,178 @@
...
+/**
+ * Tests packing of data in CDS to json
+ */
+struct WriterPackData : public ::testing::Test {
+ protected:
+  void SetUp() {}
+  void TearDown() {}
+};
+
+Data &data_ = Data::getInstance();
```
MiltFra, line 17: Also, this is dangerous. I believe it works here but the point of `getInstance` in general is to do it after initialisation (i.e. in function calls) not before. Also the naming convention is off in that case. I'd say get a reference locally or at least put it as a protected member in the test fixture structs.

MiltFra (changes requested):

test/telemetry/writer_pack_data.test.cpp
adenhaus: I asked about this last week. What must I do instead?

adenhaus: I think it wants `hyped::telemetry::` is that right?

test/telemetry/writer_pack_data.test.cpp
```diff
@@ -0,0 +1,178 @@
...
+#include "telemetry/writer.hpp"
+
+namespace hyped::testing {
+
+/**
+ * Tests packing of data in CDS to json
+ */
+struct WriterPackData : public ::testing::Test {
+ protected:
+  data::Data &data = data::Data::getInstance();
```
MiltFra, line 12: I know this is within the style guide, but I think it's even less clear now. In the `TEST_F` it looks like `data` is a local variable but it isn't. This is because of the missing underscore which we don't use for structs. Why do we not use it for structs? Because typically we use them to hold data. Nothing else. In particular, we don't usually implement methods on them (though C++ allows this, and sometimes it is required). I'd change the test fixture from `struct` to `class` and add the required underscore(s). I know this is a shortcoming in other test files, don't worry. We should fix that.

src/telemetry/sender.cpp
```diff
@@ -21,22 +21,25 @@ void Sender::run()
 {
   log_.debug("Telemetry Sender thread started");
 
-  uint16_t num_packages_sent = 0;
+  uint32_t num_packages_sent = 0;
 
   while (true) {
```
MiltFra, line 26: I keep finding things, sorry. This should be `while (sys.running_)` (or in new terminology `while (sys.isRunning())`, right? Otherwise this thread will hang potentially forever if we never enter a failure state (which we won't because other threads are exiting as intended).

MiltFra (changes requested):

src/telemetry/writer.hpp
```diff
@@ -27,14 +27,19 @@ namespace hyped::telemetry {
...
-  explicit Writer(data::Data &data);
+  explicit Writer();
 
-  // separate functions that allow to better manage data points based on their purpose
+  // functions to pack timestamp and number of packages
   void packTime();
+  void packId(const uint16_t id);
+
+  // specific functions that allow packing of structs in central data structure
   void packId(int id);
```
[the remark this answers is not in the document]
adenhaus: Can I just do `int`?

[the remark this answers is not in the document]
MiltFra: We don't like that because we are working on different architectures so giving the precise size of the type is helpful. I'm fairly confident that `int` will be identical to `int32_t`, but making it explicit is better. Also, I haven't read the standard...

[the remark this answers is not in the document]
MiltFra: Maybe also add a comment saying after how many seconds/minutes/hours/days/years we will overflow.

src/telemetry/writer.cpp
```diff
@@ -21,7 +21,9 @@ void Writer::packTime()
                         .count());
 }
 
-void Writer::packId(const int id)
+// Will overflow if 4,294,967,295 packages are sent or after about 13
+// years if we send a package every 100ms
```
MiltFra, line 25: Perfect :joy:

MiltFra (approved):

src/telemetry/writer.cpp
adenhaus: ; )

arjunnaha (approved):

PR #88, opened 2021-11-11
counting packages sent

adenhaus:
## Description
Counting json packages sent from client. Intention is to count packages received on base station side as well and make sure they match.

### DIFF
--- a/src/telemetry/sendloop.cpp
+++ b/src/telemetry/sendloop.cpp
@@ -1,8 +1,9 @@
 #include "sendloop.hpp"
 #include "writer.hpp"
-#include "data/data.hpp"
 
 #include <string>
+
+#include "data/data.hpp"
 
 namespace hyped {
 namespace telemetry {
@@ -36,7 +37,7 @@ void SendLoop::run()
 
     if (!main_ref_.client_.sendData(writer.getString())) {
       log_.ERR("Telemetry", "Error sending message");
-      telemetry_data.module_status   = data::ModuleStatus::kCriticalFailure;
+      telemetry_data.module_status = data::ModuleStatus::kCriticalFailure;
       data_.setTelemetryData(telemetry_data);
 
       break;
--- a/src/telemetry/sendloop.cpp
+++ b/src/telemetry/sendloop.cpp
@@ -1,5 +1,6 @@
 #include "sendloop.hpp"
 #include "writer.hpp"
+#include "data/data.hpp"
 
 #include <string>
 
@@ -28,14 +29,19 @@ void SendLoop::run()
     writer.packAdditionalData();
     writer.end();
 
+    data::Telemetry telemetry_data = data_.getTelemetryData();
+
     if (!main_ref_.client_.sendData(writer.getString())) {
       log_.ERR("Telemetry", "Error sending message");
-      data::Telemetry telemetry_data = data_.getTelemetryData();
       telemetry_data.module_status   = data::ModuleStatus::kCriticalFailure;
       data_.setTelemetryData(telemetry_data);
 
       break;
     }
+
+    // Increment num packages sent
+    telemetry_data.num_packages_sent += 1;
+    data_.setTelemetryData(telemetry_data);
 
     utils::concurrent::Thread::sleep(100);
   }
--- a/src/telemetry/sendloop.cpp
+++ b/src/telemetry/sendloop.cpp
@@ -19,11 +19,14 @@ void SendLoop::run()
 {
   log_.DBG("Telemetry", "Telemetry SendLoop thread started");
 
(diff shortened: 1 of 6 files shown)


MiltFra:
Do we want to have this on master?

MiltFra:
So I've just had a think about this: This is actually a good number to have and to send to the base station. We can combine it with the timestamp into a primary key if we ever want to have a database of all runs and it allows us to add a timestamp the moment we receive it to analyse delays. Timestamps are not guaranteed to be unique, especially if we're dealing in micros or even milis.

src/telemetry/sendloop.cpp
```diff
@@ -28,15 +29,20 @@ void SendLoop::run()
...
       log_.ERR("Telemetry", "Error sending message");
-      data::Telemetry telemetry_data = data_.getTelemetryData();
       telemetry_data.module_status   = data::ModuleStatus::kCriticalFailure;
       data_.setTelemetryData(telemetry_data);
 
       break;
     }
 
+    // Increment num packages sent
+    telemetry_data.num_packages_sent += 1;
```
MiltFra, line 43: I'm not sure we should write this to data. This is not useful information to anyone but the writer. Writes to data are much more expensive than local variables (or class members).

MiltFra (changes requested):

src/telemetry/sendloop.cpp
```diff
@@ -28,15 +29,20 @@ void SendLoop::run()
...
     if (!main_ref_.client_.sendData(writer.getString())) {
       log_.ERR("Telemetry", "Error sending message");
-      data::Telemetry telemetry_data = data_.getTelemetryData();
       telemetry_data.module_status   = data::ModuleStatus::kCriticalFailure;
       data_.setTelemetryData(telemetry_data);
 
       break;
     }
 
+    // Increment num packages sent
```
MiltFra, line 42: This reminds me of the textbook example of pointless comments. 

[E.g.](https://stackoverflow.blog/2021/07/05/best-practices-for-writing-code-comments/) this is called the "canonical bad example"
```
i = i + 1;         // Add one to i
```

adenhaus: Lol fair enough

src/telemetry/sendloop.cpp
adenhaus: Okay understood, I'll just pack it in with the json and send it

adenhaus:
@MiltFra what's the issue with the failing build? Local tests pass.

MiltFra (changes requested):
Please fix your auto-formatting; you're changing lines without any reason.

adenhaus:
> Please fix your auto-formatting; you're changing lines without any reason.

src/telemetry/writer.hpp
```diff
@@ -31,6 +31,7 @@ class Writer {
 
   // separate functions that allow to better manage data points based on their purpose
   void packTime();
+  void packID(uint16_t id);
```
MiltFra, line 34: should be `packId`

src/telemetry/writer.cpp
```diff
@@ -18,6 +18,12 @@ void Writer::packTime()
                         .count());
 }
 
+void Writer::packID(uint16_t id)
```
MiltFra, line 21: `const`

src/telemetry/sendloop.cpp
```diff
@@ -18,25 +20,31 @@ void SendLoop::run()
...
-      data::Telemetry telemetry_data = data_.getTelemetryData();
-      telemetry_data.module_status   = data::ModuleStatus::kCriticalFailure;
+      telemetry_data.module_status = data::ModuleStatus::kCriticalFailure;
       data_.setTelemetryData(telemetry_data);
 
       break;
     }
 
+    num_packs_sent += 1;
+
```
MiltFra, line 47: there's lots of whitespaces around; wondering whether that makes sense? I know you're just fitting your changes into the style that was already there so I suppose it's fine.

src/telemetry/sendloop.cpp
```diff
@@ -18,25 +20,31 @@ void SendLoop::run()
 {
   log_.DBG("Telemetry", "Telemetry SendLoop thread started");
 
+  uint16_t num_packs_sent = 0;
```
MiltFra, line 23: I'd call it `num_packages_sent` but idk, won't stop me from approving so dw

MiltFra (changes requested):

MiltFra (approved):

sashank17 (approved):

adenhaus:
Hi @MiltFra please merge thanks boss

PR #91, opened 2021-11-12
Replaced default with individual states

amadrafi:
I replaced all the defaults statements and for brakes, I wasn't sure about the critical failure function so I put navigation as a placeholder for kInvalid.

### DIFF
--- a/src/brakes/main.cpp
+++ b/src/brakes/main.cpp
@@ -115,7 +115,15 @@ void Main::run()
           f_brake_->checkAccFailure();
         }
         break;
-      default:
+
+      //Default for all the other states
+      case data::State::kReady:
+      case data::State::kFailureStopped:
+        break;
+
+      case data::State::kInvalid:
+        brakes_               = data_.getEmergencyBrakesData();
+        brakes_.module_status = ModuleStatus::kCriticalFailure;
         break;
     }
   }
--- a/src/brakes/main.cpp
+++ b/src/brakes/main.cpp
@@ -115,12 +115,9 @@ void Main::run()
           f_brake_->checkAccFailure();
         }
         break;
-
-      //Default for all the other states
       case data::State::kReady:
       case data::State::kFailureStopped:
         break;
-
       case data::State::kInvalid:
         brakes_               = data_.getEmergencyBrakesData();
         brakes_.module_status = ModuleStatus::kCriticalFailure;
--- a/src/brakes/main.cpp
+++ b/src/brakes/main.cpp
@@ -87,6 +87,7 @@ void Main::run()
         break;
       case data::State::kNominalBraking:
       case data::State::kEmergencyBraking:
+      case data::State::kInvalid:
         if (!m_brake_->checkClamped()) { m_brake_->sendClamp(); }
         if (!f_brake_->checkClamped()) { f_brake_->sendClamp(); }
         Thread::sleep(data::EmergencyBrakes::kBrakeCommandWaitTime);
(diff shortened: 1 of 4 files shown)


src/sensors/gpio_manager.cpp
```diff
@@ -94,7 +94,9 @@ void GpioManager::run()
           setHP();
           log_.INFO("GPIO-MANAGER", "kReady...HP SSR set and HP on");
           break;
-        default:      // undefied behaviour, e.g. kInvalid
+
+        //Default for all the other states
+        case data::State::kInvalid:      // undefied behaviour, e.g. kInvalid
```
MiltFra, line 99: Remove the `e.g. kInvalid`

src/propulsion/main.cpp
```diff
@@ -73,7 +73,11 @@ void Main::run()
       case data::State::kEmergencyBraking:
         state_processor_->quickStopAll();
         break;
-      default:
+
```
MiltFra, line 76: Remove the empty line

src/navigation/main.cpp
```diff
@@ -55,9 +55,14 @@ void Main::run()
         nav_.navigate();
         break;
 
-      default:
+      //Default for all the other states
+      case State::kFailureStopped:
+      case State::kFinished:
         navigation_complete = true;
         break;
+
```
MiltFra, line 63: empty line

src/navigation/main.cpp
```diff
@@ -55,9 +55,14 @@ void Main::run()
         nav_.navigate();
         break;
 
-      default:
+      //Default for all the other states
```
MiltFra, line 58: i don't think we need this comment; it only makes sense in this context, i.e. it won't give any insight in the future

src/brakes/main.cpp
```diff
@@ -115,7 +115,15 @@ void Main::run()
           f_brake_->checkAccFailure();
         }
         break;
-      default:
+
+      //Default for all the other states
```
MiltFra, line 119: ...

MiltFra (changes requested):
Looking at the warnings generated by the compiler, I've also realised that we need to move `num_states` out of the enum into a constexpr `kNumStates`. This will require you to touch quite a few files.

src/brakes/main.cpp
```diff
@@ -115,7 +115,12 @@ void Main::run()
           f_brake_->checkAccFailure();
         }
         break;
-      default:
+      case data::State::kReady:
+      case data::State::kFailureStopped:
+        break;
+      case data::State::kInvalid:
+        brakes_               = data_.getEmergencyBrakesData();
+        brakes_.module_status = ModuleStatus::kCriticalFailure;
```
MiltFra, line 123: You need to write back to data to set critical failure. Also, I'd assume we want to handle this similarly to `kEmergencyBraking`.

src/navigation/main.cpp
```diff
@@ -54,10 +54,12 @@ void Main::run()
...
         nav_.navigate();
         break;
-
-      default:
+      case State::kFailureStopped:
+      case State::kFinished:
         navigation_complete = true;
         break;
+      case State::kInvalid:
+        break;
```
MiltFra, line 62: We probably don't want to do nothing here. Either we stop navigation to free up processing to make sure we can shut down properly or we keep navigating to keep reporting displacement/velocity/acceleration.

MiltFra (changes requested):

SnickeyX (approved):

MiltFra (approved):

PR #104, opened 2022-01-24
changed default to not use keyence

maxguy2001:
Changed navigation system to not compare keyence to other measurements by default


callumTM (approved):
very good

PR #107, opened 2022-01-27
Arduino code for IR sensor on Lateral Suspension

supermegho:
## Description
This is the initial Arduino code for counting the wheel revolutions on a lateral suspension wheel. A signal is sent to the BBB every time an input is received by one the IR pins.

 *Click up link* [](<link here>)

## Changes
- IR sensor pin takes an input
- Output is sent to BBB.

## Additional Info 
screenshots, commands to run etc.

### DIFF
--- a/arduino/irsensor.cpp
+++ b/arduino/irsensor.cpp
@@ -23,42 +23,34 @@ void setup(){
 }
 // Read IR Sensors + Output based on reading
 void loop(){
-    statusSensor = digitalRead (IRSensor1);
+    statusSensor1 = digitalRead (IRSensor1);
+    statusSensor2 = digitalRead (IRSensor2);
+    statusSensor3 = digitalRead (IRSensor3);
+    statusSensor4 = digitalRead (IRSensor4);
 
-    if (statusSensor == 1){
+    if (statusSensor1 == 1){
         digitalWrite (Output1, LOW);
     }
     else {
         digitalWrite (Output1, HIGH);
     }
-}
-void loop(){
-    statusSensor = digitalRead (IRSensor2);
-
-    if (statusSensor == 1){
+    
+    if (statusSensor2 == 1){
         digitalWrite (Output2, LOW);
     }
     else {
         digitalWrite (Output, HIGH);
     }
-}
-void loop(){
-    statusSensor = digitalRead (IRSensor3);
-
-    if (statusSensor == 1){
-        digitalWrite (Output3, LOW);
+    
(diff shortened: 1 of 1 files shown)


arduino/irsensor.cpp
```diff
@@ -0,0 +1,56 @@
...
+    }
+    else {
+        digitalWrite (Output, HIGH);
+    }
+    if (statusSensor4 == 1){
+        digitalWrite (Output4, LOW);
+    }
+    else {
+        digitalWrite (Output4, HIGH);
+    }
```
MiltFra, line 56: This is missing a closing bracket, no?

arduino/irsensor.cpp
```diff
@@ -0,0 +1,56 @@
...
+void setup(){
+    // IRSensors are inputs
+    pinMode (IRSensor1, INPUT);
+    pinMode (IRSensor2, INPUT);
+    pinMode (IRSensor3, INPUT);
+    pinMode (IRSensor4, INPUT);
+    // Outputs
+    pinMode (Output1, OUTPUT);
+    pinMode (Output2, OUTPUT);
+    pinMode (Output3, OUTPUT);
```
MiltFra, line 21: Formatting. `pinMode(Output3, OUTPUT);`

arduino/irsensor.cpp
```diff
@@ -0,0 +1,56 @@
...
+    pinMode (Output4, OUTPUT);
+}
+// Read IR Sensors + Output based on reading
+void loop(){
+    statusSensor1 = digitalRead (IRSensor1);
+    statusSensor2 = digitalRead (IRSensor2);
+    statusSensor3 = digitalRead (IRSensor3);
+    statusSensor4 = digitalRead (IRSensor4);
+
+    if (statusSensor1 == 1){
```
MiltFra, line 31: No need to check the `== 1`. Just do `if (statusSensor1) { ... }`. Looks more intuitive, I think.

MiltFra (changes requested):
This just passes on the signals, right? I'd like to see the counting implemented before we pass this on. @kshxtij

arduino/irsensor.cpp
supermegho: Yes, you are correct

SnickeyX:
@kshxtij and I spent some time trying to get the arduino to work. Most things failed, the strings method works for now; so committing this while we keep experimenting.

kshxtij:
Replaced by #120.

PR #109, opened 2022-01-31
Refactor telemetry tests

adenhaus:
Deleted the variables that held error strings and put the strings inline for existing telemetry tests.

### DIFF
--- a/test/telemetry/writer_convert_module_status.test.cpp
+++ b/test/telemetry/writer_convert_module_status.test.cpp
@@ -11,12 +11,6 @@ using namespace hyped::data;
  * Tests the conversion of Module Status from Enum to String as required for GUI
  */
 struct WriterConvertModuleStatus : public ::testing::Test {
-  // ---- Error messages -------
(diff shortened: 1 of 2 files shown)


MiltFra (changes requested):
Looking good. While you're at it, you may want to add a test (and possibly support) for `kPreCalibrating`.

src/telemetry/main.cpp
```diff
@@ -56,4 +55,4 @@ void Main::run()
   log_.DBG("Telemetry", "Exiting Telemetry Main thread");
 }
 
-}  // namespace hyped
+}  // namespace telemetry
```
MiltFra, line 58: I think this should be done by auto formatting? It should be `hyped::telemetry` I believe.

MiltFra (approved):

arjunnaha (approved):
LGTM!

weijuinlee (approved):
Looks alright. What's was the motive behind doing error handling inline?

PR #110, opened 2022-02-07
Nav: navigation.cpp/hpp testing

callumTM:
## Description
Testing for `navigation.cpp/hpp`, currently being used to try and understand the system better


test/navigation/navigation.test.cpp
```diff
@@ -0,0 +1,35 @@
...
+  hyped::navigation::Navigation nav(log_, 0);
+
+  // unsure of how or what to set the sensors data to
+  Randomiser::randomiseSensorsData(sensors_data_);
+
+  data_.setSensorsData(sensors_data_);
+
+  // this should read the sensors data and use it to update velocity etc. but doesn't seem to
+  nav.navigate();
+  std::cout << data_.getNavigationData().velocity
```
maxguy2001, line 33: semicolon at the end of this line?

test/navigation/navigation.test.cpp
```diff
@@ -0,0 +1,35 @@
+#include "randomiser.hpp"
+#include "test.hpp"
+
+#include <iostream>
+#include <string>
+
+#include "gtest/gtest.h"
```
maxguy2001, line 7: '#include <gtest/gtest.h>'?

MiltFra: This *may* not be an issue but it is highly advised as it's part of a different binary (to some degree, let's not get into the details of this). Essentially these headers are clearly not local to the source file.

test/navigation/navigation.test.cpp
```diff
@@ -0,0 +1,35 @@
...
+  // Run before each test
+  void SetUp() {}
+  void TearDown() {}
+};
+
+namespace hyped::navigation {
+TEST_F(NavigationTest, correctEmergencyBrakingDistance)
+{
+  Data &data_ = Data::getInstance();
+  Sensors sensors_data_;
```
MiltFra, line 23: Why the underscore here? It's a local variable.

test/navigation/navigation.test.cpp
```diff
@@ -0,0 +1,35 @@
...
+  hyped::navigation::Navigation nav(log_, 0);
+
+  // unsure of how or what to set the sensors data to
+  Randomiser::randomiseSensorsData(sensors_data_);
```
MiltFra, line 27: This is fine, don't worry.

test/navigation/navigation.test.cpp
```diff
@@ -0,0 +1,35 @@
...
+
+  data_.setSensorsData(sensors_data_);
+
+  // this should read the sensors data and use it to update velocity etc. but doesn't seem to
```
MiltFra, line 31: Have you checked what the randomiser actually writes into data? Maybe the values are such that we would deterministically expect a certain velocity?

MiltFra (changes requested):

test/navigation/navigation.test.cpp
MiltFra: That's why you should put in hand-picked values because you know what to expect, by the way.

PR #111, opened 2022-02-08
Removed using keywords from all tests and changing structs to classes

adenhaus:
Removed using keywords from all tests and changed all test structs to classes


MiltFra (approved):
While we're at it, would you mind changing the test fixture to a class? Maybe even improve the other tests like that.

MiltFra (approved):

callumTM (approved):
Looks all good!

PR #116, opened 2022-02-13
Refactored outlier detection function

maxguy2001:
Refactored outlier detection function to clean up code and remove all references to "tukeyFences" to clarify functionality.

Also created an array containing the quartiles of the reliable IMU data and moved the section of code which calculates the quartiles to it's own separate function for easier testing.

### DIFF
--- a/src/navigation/navigation.cpp
+++ b/src/navigation/navigation.cpp
@@ -210,7 +210,7 @@ void Navigation::queryImus()
   log_.debug("Raw acceleration values: %.3f, %.3f, %.3f, %.3f", raw_acceleration_moving[0],
              raw_acceleration_moving[1], raw_acceleration_moving[2], raw_acceleration_moving[3]);
   // Run outlier detection on moving axis
-  imuOutlierDetection(raw_acceleration_moving, kIqrScaler);
+  (raw_acceleration_moving, kIqrScaler);
   // TODO(Justus) how to run outlier detection on non-moving axes without affecting "reliable"
   // Current idea: outlier function takes reliability write flag, on hold until z-score impl.
 
@@ -332,7 +332,7 @@ void Navigation::imuOutlierDetection(NavigationArray &data_array, const data::na
   
   // find the thresholds
   // clip IQR to upper bound to avoid issues with very large outliers
-  const auto iqr         = std::min(q3 - q1, kMinIQRBound);
+  const auto iqr         = std::min(q3 - q1, kMaxInterQuartileRange);
   const auto upper_limit = q3 + threshold * iqr;
   const auto lower_limit = q1 - threshold * iqr;
   // replace any outliers with the median
--- a/src/navigation/navigation.cpp
+++ b/src/navigation/navigation.cpp
@@ -308,7 +308,7 @@ void Navigation::logWrite()
   write_to_file_ = true;
 }
 
-void Navigation::imuOutlierDetection(NavigationArray &data_array, const data::nav_t threshold)
+void Navigation::calculateImuQuartiles(NavigationArray &data_array)
 {
   std::vector<data::nav_t> data_vector;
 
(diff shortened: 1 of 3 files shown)


src/navigation/navigation.cpp
```diff
@@ -307,60 +308,31 @@ void Navigation::logWrite()
...
+    q2 = data_vector.at(1);
+  } else {
+    status_ = data::ModuleStatus::kCriticalFailure;
+    log_.error("At least two IMUs no longer reliable, entering CriticalFailure.");
+  }
+  
   // find the thresholds
   // clip IQR to upper bound to avoid issues with very large outliers
-  const auto iqr         = std::min(q3 - q1, kTukeyIQRBound);
+  const auto iqr         = std::min(q3 - q1, kMinIQRBound);
```
MiltFra, line 335: Please explain.

MiltFra: Two things: (a) you should absolutely rename it to something involving max or upper because the value on the left cannot be greater. (b) I don't like the usage of the abbreviation IQR (here and elsewhere). I think it stands for intermediate quartile ranges but I am not sure, really. This is a sign that maybe we should rename it.

src/navigation/navigation.cpp
```diff
@@ -307,60 +308,31 @@ void Navigation::logWrite()
...
+
+  for (size_t i = 0; i < data::Sensors::kNumImus; ++i){
+    if (is_imu_reliable_.at(i)) {
+      data_vector.push_back(data_array.at(i));
     }
-    // do not check for further outliers because no reliable detection could be made!
-    return;
   }
+  std::sort(data_vector.begin(), data_vector.end());
+  data::nav_t q1, q2, q3;
```
MiltFra, line 321: Actually, this should probably be `std::array<data::nav_t, 3> quartile_bounds`. Now that I think about it (you see, this is all very WIP), why don't we create a function, e.g. `calculate_imu_quartile_bounds`, that takes the IMU readings and a mask and returns the quartiles? We could then unit test it right away.

src/navigation/navigation.hpp
```diff
@@ -159,6 +165,8 @@ class Navigation {
   std::array<data::nav_t, 3> calibration_limits_;
   // Calibration variances in each dimension, necessary for vibration checking
   std::array<data::nav_t, 3> calibration_variance_;
+  // Quartiles of reliable IMU readings of form (q1, q2(\median), q3)
+  std::array<data::nav_t, 3> quartile_bounds;
```
MiltFra, line 169: Two issues with this. (a) the naming is wrong again which lead me to search for the local declaration and (b) making this a class member is spaghetti code. Declare it locally in `calculateImuQuartiles` and then return it (you know, functions can have return types other than `void`). Then you can do `const auto quartile_bound = calculateImuQuartiles(...)` in the outlier detection method.

src/navigation/navigation.hpp
```diff
@@ -136,8 +142,8 @@ class Navigation {
 
   static constexpr int kPrintFreq                     = 1;
   static constexpr data::nav_t kEmergencyDeceleration = 24;
-  static constexpr data::nav_t kIqrScaler        = 1;  // 0.75
-  static constexpr data::nav_t kMinIQRBound         = 3;
+  static constexpr data::nav_t kIqrScaler             = 1;  // 0.75
```
MiltFra, line 145: There is an IQR again...

src/navigation/navigation.cpp
```diff
@@ -307,69 +308,45 @@ void Navigation::logWrite()
...
-    // TODO(Sury): Handle possible division by zero if all data points are 0.0 or explain why this
-    // can't happen.
-    for (data::nav_t &imu_data : data_array) {
-      imu_data = sum_non_outliers / num_non_outliers;
-    }
-    // do not check for further outliers because no reliable detection could be made!
-    return;
+    quartile_bounds.at(1) = data_vector.at(1);
+  } else {
+    status_ = data::ModuleStatus::kCriticalFailure;
```
MiltFra, line 328: I think we should write to `data` here right away.

src/navigation/navigation.cpp
```diff
@@ -202,14 +203,14 @@ void Navigation::queryImus()
...
-    // the moving axis should be set to 0 for tukeyFences
+    // the moving axis should be set to 0 for outlier detection
     if (!is_imu_reliable_.at(i)) { raw_acceleration_moving.at(i) = 0; }
     raw_acceleration_moving.at(i) = acceleration[movement_axis_];
   }
   log_.debug("Raw acceleration values: %.3f, %.3f, %.3f, %.3f", raw_acceleration_moving[0],
              raw_acceleration_moving[1], raw_acceleration_moving[2], raw_acceleration_moving[3]);
   // Run outlier detection on moving axis
-  tukeyFences(raw_acceleration_moving, kTukeyThreshold);
+  (raw_acceleration_moving, kInterQuartileScaler);
```
MiltFra, line 213: This looks wrong.

src/navigation/navigation.cpp
```diff
@@ -307,69 +308,47 @@ void Navigation::logWrite()
...
 
-void Navigation::tukeyFences(NavigationArray &data_array, const data::nav_t threshold)
+Navigation::quartile_bounds Navigation::calculateImuQuartiles(NavigationArray &data_array)
 {
-  // Define the quartiles first:
-  data::nav_t q1 = 0;
-  data::nav_t q2 = 0;
-  data::nav_t q3 = 0;
-  // The most likely case is that all four IMUs are still reliable:
+  std::array<data::nav_t, 3> quartile_bounds;
```
MiltFra, line 313: Move this down to be closer to its first use.

src/utils/io/spi.cpp
```diff
@@ -266,6 +266,6 @@ SPI::~SPI()
 
   close(spi_fd_);
 }
-}  // namespace io
 }  // namespace utils
 }  // namespace hyped
+}  // namespace hyped
```
MiltFra, line 271: something is off...

src/navigation/navigation.hpp
```diff
@@ -136,8 +145,8 @@ class Navigation {
 
   static constexpr int kPrintFreq                     = 1;
   static constexpr data::nav_t kEmergencyDeceleration = 24;
-  static constexpr data::nav_t kTukeyThreshold        = 1;  // 0.75
-  static constexpr data::nav_t kTukeyIQRBound         = 3;
+  static constexpr data::nav_t kInterQuartileScaler   = 1;  // 0.75
```
MiltFra, line 148: confusing comment

src/navigation/navigation.hpp
```diff
@@ -27,6 +27,7 @@ class Navigation {
   using NavigationArray          = std::array<data::nav_t, data::Sensors::kNumImus>;
   using NavigationArrayOneFaulty = std::array<data::nav_t, data::Sensors::kNumImus - 1>;
   using FilterArray              = std::array<KalmanFilter, data::Sensors::kNumImus>;
+  using quartile_bounds          = std::array<data::nav_t, 3>;
```
MiltFra, line 30: Type naming...

MiltFra (changes requested):
Overall pretty good stuff; some polishing still required, though.

src/utils/io/spi.cpp
SnickeyX: this is still there

src/navigation/navigation.cpp
```diff
@@ -307,69 +308,48 @@ void Navigation::logWrite()
...
-    // TODO(Sury): Handle possible division by zero if all data points are 0.0 or explain why this
-    // can't happen.
-    for (data::nav_t &imu_data : data_array) {
-      imu_data = sum_non_outliers / num_non_outliers;
-    }
-    // do not check for further outliers because no reliable detection could be made!
-    return;
+    quartile_bounds.at(1) = data_vector.at(1);
+  } else {
+    auto navigation_data          = data_.getNavigationData();
```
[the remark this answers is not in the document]
MiltFra: no it can't

adenhaus (approved):

src/utils/io/spi.cpp
MiltFra: bump

kshxtij (approved):
Can be merged after the namespace comment is fixed.

MiltFra (approved):

PR #117, opened 2022-02-13
Tel gui fix

adenhaus:
Fixing issues that were affecting GUI compiling and running


src/sensors/fake_batteries.cpp
```diff
@@ -10,9 +10,9 @@ FakeBatteries::FakeBatteries(Logger &log, bool is_lp, bool is_fail)
     : data_(Data::getInstance()),
       log_(log),
       lp_failure_{{100, 1000, 0, 0, 100, 0, 0, 0}},
-      lp_success_{{245, 450, 70, 0, 35, 0, 0, 0}},
+      lp_success_{{35, 35, 35, 35, 35, 35, 35}},
       hp_failure_{{900, 10000, 0, 5, 70, 100, 500, 400}},
-      hp_success_{{1150, 1000, 70, 30, 35, 40, 3000, 3200}},
+      hp_success_{{35, 35, 35, 35, 35, 35, 35}},
```
MiltFra, line 15: @kshxtij can you please confirm that these values are sensible? They don't need to be perfect, just on the right order of magntitude.

kshxtij: - Voltage is measured in mV https://github.com/Hyp-ed/hyped-2022/blob/324e24d5fb310a19e7d04d7946038d03304e1017/src/sensors/bms.cpp#L135 so the hp_success value should be between 3300 and 4200 imo, rest looks right.

MiltFra: @adenhaus do those values do the trick? If not, we need to figure out what's going on.

adenhaus: @arjunnaha is the one working on this

adenhaus: @kshxtij when I change `hp_success` values to 3500 I get following errors:

`14:47:34.517 ERROR[BMS-MANAGER]: BMS HP 0 voltage out of range: 3500`
`14:47:34.517 ERROR[BMS-MANAGER]: battery failure detected`
`14:47:34.518 ERROR[STATE-MACHINE]: critical failure in batteries`

adenhaus: Okay @kshxtij says 35 is correct after all : )

kshxtij (approved):

configurations/config.json
```diff
@@ -1,28 +1,83 @@
...
+      86
+    ],
     "temperature_pin": 3
   },
   "telemetry": {
     "server_ip": "127.0.01",
     "server_port": "9090"
-  }
+  },
+  "fake_trajectory": {
```
MiltFra, line 42: This config is intended for the run so please remove all references to the fake system here. We need to be able to catch faulty values and adding a bunch of nonsense isn't going to help with that.

configurations/config.server.json
```diff
@@ -0,0 +1 @@
+{"system":{"use_fake_brakes":false,"use_fake_high_power":false,"use_fake_temperature_fail":false,"use_fake_controller":false,"log_level":1,"use_fake_batteries":false,"use_fake_temperature":false,"use_fake_batteries_fail":false,"use_fake_trajectory":false,"axis":0},"sensors":{"keyence_pins":[67,68],"imu_pins":[22,27,47,86],"bms_startup_time_micros":5000000,"temperature_pin":3},"brakes":{"button_pins":[66,69],"command_pins":[87,10]},"telemetry":{"server_ip":"127.0.01","server_port":"9090"}}
```
MiltFra, line 1: This still needs to be removed after you've added it to the .gitignore.

MiltFra (changes requested):
Just small things

MiltFra (approved):

PR #120, opened 2022-03-03
Serial Communication

kshxtij:
## Description
Implements UART communication with the Arduino to receive structs containing wheel encoder data.

## Implemented

- [x] Updated Arduino Code
- [x] Serial Interface in hyped::utils::io

## ToDo

- [ ] Implement receiver
- [ ]  Add struct to central data structure

### DIFF
--- a/arduino/main.ino
+++ b/arduino/main.ino
@@ -1,0 +1,49 @@
+#include <SoftwareSerial.h>
+#include <SerialProtocol.h>
+#include <StreamSerialProtocol.h>
+#include <ArduinoSerialProtocol.h>
+
+int IRSensor1 = 2;
+int IRSensor2 = 3;
+int IRSensor3 = 4;
+int IRSensor4 = 5;
+
+// The payload that will be sent to the other device
+struct Payload {
+    int16_t wheelEncoderA;
+    int16_t wheelEncoderB;
+    int16_t wheelEncoderC;
+    int16_t wheelEncoderD;
+} payload;
+
+ArduinoSerialProtocol protocol(&Serial, (uint8_t*)&payload, sizeof(payload));
+uint8_t receiveState;
+
+void setup() 
+{
+    Serial.begin(9600);
+
+    payload.wheelEncoderA = 0;
+    payload.wheelEncoderB = 0;
+    payload.wheelEncoderC = 0;
+    payload.wheelEncoderD = 0;
+}
+
+void loop()
+{
+    if(digitalRead(IRSensor1)){
+        ++payload.wheelEncoderA;
+    }
+    if(digitalRead(IRSensor2)){
+        ++payload.wheelEncoderB;
+    }
+    if(digitalRead(IRSensor3)){
+        ++payload.wheelEncoderC;
+    }
+    if(digitalRead(IRSensor4)){
+        ++payload.wheelEncoderD;
+    }
+
+    protocol.send();
+}
+
--- a/src/utils/io/serial.cpp
+++ b/src/utils/io/serial.cpp
@@ -1,0 +1,138 @@
+#include "serial.h"
+
+#include <fcntl.h>
+#include <stdio.h>
+#include <time.h>
+#include <unistd.h>
+
+#define HEADER 'S'
+#define BYTES_TO_PAYLOAD 2
+
+/* Message format:
+    Header       (1 byte)            - Header Byte
+    DataLength   (1 byte)            - The size of the payload
+    Data         (DataLength bytes)  - Actual payload
+    Checksum     (1 byte)            - Message integrity check
+
+    Total message size = (payload size + 3) bytes;
+*/
+
+namespace hyped::utils::io {
+
+SerialProtocol::SerialProtocol(int serial, uint8_t *payload, uint8_t payloadSize)
+{
+  setSerial(serial);
+  this->payload     = payload;
+  this->payloadSize = payloadSize;
+  inputBuffer       = (uint8_t *)malloc(payloadSize);
+}
+
+void SerialProtocol::setSerial(int serial)
+{
+  this->serial = serial;
+};
+
+uint8_t SerialProtocol::send()
+{
+  if (!serialAvailable()) {
+    // There is no place to send the data
+    return ProtocolState::NO_SERIAL;
+  }
+
+  uint8_t checksum = payloadSize;
+
+  // Write the HEADER
+  sendData(HEADER);
+  sendData(payloadSize);
+
+  // Write the payload
+  for (int i = 0; i < payloadSize; i++) {
+    checksum ^= *(payload + i);
+    sendData(*(payload + i));
+  }
+
+  // Complete with the checksum
+  sendData(checksum);
+
+  return ProtocolState::SUCCESS;
+}
+
+// This method is blocking while there is data in the serial buffer
+uint8_t SerialProtocol::receive()
+{
+  uint8_t data;
+
+  if (!serialAvailable()) {
+    // Cannot get any data
+    return ProtocolState::NO_SERIAL;
+  }
+
+  while (true) {
+    data = readData();
+
+    if (data == -1) {
+      return bytesRead > 0 ? ProtocolState::WAITING_FOR_DATA : ProtocolState::NO_DATA;
+    }
+
+    if (bytesRead == 0) {
+      // Look for HEADER
+      if (data != HEADER) { return ProtocolState::NO_DATA; }
+
+      ++bytesRead;
+      continue;
+    }
+
+    if (bytesRead == 1) {
+      // Look for payload size
+      if (data != payloadSize) {
+        bytesRead = 0;
+        return ProtocolState::INVALID_SIZE;
+      }
+
+      // The checksum starts with the payload size
+      actualChecksum = data;
+      ++bytesRead;
+      continue;
+    }
+
+    // The > check ensures that we don't overflow and
+    // that we discard bad messages
+    if (bytesRead >= (payloadSize + BYTES_TO_PAYLOAD)) {
+      // We are done with the message. Regardless of the outcome
+      // only a new message can come next
+      bytesRead = 0;
+
+      // Before we can be fully done, we need the
+      // checksum to match
+      if (actualChecksum != data) { return ProtocolState::INVALID_CHECKSUM; }
+
+      memcpy(payload, inputBuffer, payloadSize);
+      return ProtocolState::SUCCESS;
+    }
+
+    inputBuffer[bytesRead - BYTES_TO_PAYLOAD] = data;
+    actualChecksum ^= data;
+
+    ++bytesRead;
+  }
+}
+
+bool SerialProtocol::serialAvailable()
+{
+  return serial != -1;
+}
+
+void SerialProtocol::sendData(uint8_t data)
+{
+  write(serial, &data, 1);
+}
+
+uint8_t SerialProtocol::readData()
+{
+  uint8_t data[1];
+  int cnt = read(serial, (void *)data, 1);
+  if (cnt == 0) { return -1; }
+
+  return data[0];
+}
+};  // namespace hyped::utils::io
--- a/src/utils/io/serial.h
+++ b/src/utils/io/serial.h
@@ -1,0 +1,61 @@
+#include <stdlib.h>
+#include <string.h>
+#include <stdint.h>
+
+namespace hyped::utils::io {
+
+struct ProtocolState {
+    enum Enum 
+    {
+        // The serial object was not set
+        NO_SERIAL = 0,
+
+         // The operation succeeded
+        SUCCESS = 1,
+
+        // There is not (enough) data to process
+        NO_DATA = 2,
+        
+         // The object is being received but the buffer doesn't have all the data
+        WAITING_FOR_DATA = 3,
+
+        // The size of the received payload doesn't match the expected size
+        INVALID_SIZE = 4,
+
+        // The object was received but it is not the same as one sent
+        INVALID_CHECKSUM = 5
+    };
+};
+
+class SerialProtocol {
+    public:
+        SerialProtocol(int, uint8_t*, uint8_t);
+
+        // Sends the current payload
+        
+        // Returns a ProtocolState enum value
+        uint8_t send();
+
+        // Tries to receive the payload from the
+        // current available data
+        // Will replace the payload if the receive succeeds
+
+        // Returns a ProtocolState enum value
+        uint8_t receive();
+        
+        void setSerial(int);
+        bool serialAvailable();
+        void sendData(uint8_t);
+        uint8_t readData();
+
+    private:
+        int serial;
+        uint8_t* payload;
+        uint8_t payloadSize;
+
+        uint8_t* inputBuffer;
+        uint8_t bytesRead;
+
+        uint8_t actualChecksum;
+    };
+};


src/utils/io/serial.cpp
```diff
@@ -0,0 +1,138 @@
...
+{
+  uint8_t data;
+
+  if (!serialAvailable()) {
+    // Cannot get any data
+    return ProtocolState::NO_SERIAL;
+  }
+
+  while (true) {
+    data = readData();
```
MiltFra, line 71: Declare `data` here, not outside.

src/utils/io/serial.cpp
```diff
@@ -0,0 +1,138 @@
...
+
+    if (data == -1) {
+      return bytesRead > 0 ? ProtocolState::WAITING_FOR_DATA : ProtocolState::NO_DATA;
```
MiltFra, line 74: Use an if statement and not ternary. Difficult to read.

src/utils/io/serial.cpp
```diff
@@ -0,0 +1,138 @@
...
+
+void SerialProtocol::sendData(uint8_t data)
+{
+  write(serial, &data, 1);
+}
+
+uint8_t SerialProtocol::readData()
+{
+  uint8_t data[1];
+  int cnt = read(serial, (void *)data, 1);
```
MiltFra, line 133:
```cpp
uint8_t data;
int cnt = read(serial, (void *)(&data), 1);
```

src/utils/io/serial.hpp
```diff
@@ -0,0 +1,61 @@
+#include <stdlib.h>
+#include <string.h>
+#include <stdint.h>
+
+namespace hyped::utils::io {
+
+struct ProtocolState {
+    enum Enum
```
MiltFra, line 8:
```cpp
enum class ProtocolState {
   kNoSerial = 0,
   // and so on
}
```

src/utils/io/serial.hpp
```diff
@@ -0,0 +1,61 @@
...
+        
+        // Returns a ProtocolState enum value
+        uint8_t send();
+
+        // Tries to receive the payload from the
+        // current available data
+        // Will replace the payload if the receive succeeds
+
+        // Returns a ProtocolState enum value
+        uint8_t receive();
```
MiltFra, line 44: Why is the return type `uint8_t` if the comment says it should be `ProtocolState`?

src/utils/io/serial.hpp
```diff
@@ -0,0 +1,61 @@
...
+        
+        void setSerial(int);
+        bool serialAvailable();
+        void sendData(uint8_t);
```
MiltFra, line 48: Please give the arguments names for readability.

src/utils/io/serial.cpp
```diff
@@ -0,0 +1,138 @@
...
+uint8_t SerialProtocol::receive()
+{
+  uint8_t data;
+
+  if (!serialAvailable()) {
+    // Cannot get any data
+    return ProtocolState::NO_SERIAL;
+  }
+
+  while (true) {
```
MiltFra, line 70: While `sys_.is_running()` please. We don't want this to stop us from shutting the application down.

src/utils/io/serial.cpp
```diff
@@ -0,0 +1,138 @@
...
+  uint8_t checksum = payloadSize;
+
+  // Write the HEADER
+  sendData(HEADER);
+  sendData(payloadSize);
+
+  // Write the payload
+  for (int i = 0; i < payloadSize; i++) {
+    checksum ^= *(payload + i);
+    sendData(*(payload + i));
```
MiltFra, line 51: `payload[i]`?

src/utils/io/serial.cpp
MiltFra: I mean obviously the code is still bad but it's slightly less so.

src/utils/io/serial.cpp
```diff
@@ -0,0 +1,138 @@
...
+*/
+
+namespace hyped::utils::io {
+
+SerialProtocol::SerialProtocol(int serial, uint8_t *payload, uint8_t payloadSize)
+{
+  setSerial(serial);
+  this->payload     = payload;
+  this->payloadSize = payloadSize;
+  inputBuffer       = (uint8_t *)malloc(payloadSize);
```
MiltFra, line 27: Please, this is C++... You could at least have had the decency to use `new` for this. 

There are several things wrong with this. (a) even in C, every `malloc` needs a `free` otherwise you're just leaking memory, (b) from reading this the right way of doing it would be to have the `input_buffer_` (notice the naming :angry:) be a `std::vector<uint8_t>` which we then extend to the right size here.

I hate to admit it, but this calls for templates... if you add a template argument T (which is the payload) you could stop all dynamic allocations and just have an argument `buffer` which would, by design, already have the right size. Then you could make many things much easier...

src/utils/io/serial.hpp
```diff
@@ -0,0 +1,61 @@
+#include <stdlib.h>
```
MiltFra, line 1: Missing `#pragma once`

src/utils/io/serial.cpp
```diff
@@ -0,0 +1,138 @@
...
+
+    Total message size = (payload size + 3) bytes;
+*/
+
+namespace hyped::utils::io {
+
+SerialProtocol::SerialProtocol(int serial, uint8_t *payload, uint8_t payloadSize)
+{
+  setSerial(serial);
+  this->payload     = payload;
```
MiltFra, line 25: You see, if you had followed the naming conventions, you could do the following :^)

```cpp
payload_ = payload;
```

However, it would be even better to do this on initialisation. I.e. 

```cpp
SerialProtocol::SerialProtocol(const int serial, const uint8_t *payload, const uint8_t payload_size) 
   : serial_(serial), payload_(payload), input_buffer(payload_size)  
{ /*...*/ }
```
I took the liberty of fixing the naming, following my advice on vectors below and adding `const`s. So should you.

src/utils/io/serial.cpp
```diff
@@ -0,0 +1,138 @@
+#include "serial.hpp"
+
+#include <fcntl.h>
+#include <stdio.h>
+#include <time.h>
+#include <unistd.h>
+
+#define HEADER 'S'
+#define BYTES_TO_PAYLOAD 2
```
MiltFra, line 9: `static constexpr std::size_t kBytedToPayload`

MiltFra (changes requested):
Even if you hadn't violated every single rule in our style guide, I would still be advising you to rewrite all of this. I'm not sure this is good C code, but it's definitely not good C++. 

You are using `malloc`, plenty of raw pointers, `int` as file descriptors, and many other things that don't have a place here. To solve the problem in question, I advise you to have a look at the following: 

- https://www.cplusplus.com/reference/istream/istream/operator%3E%3E/
- https://www.cplusplus.com/reference/fstream/ifstream/
- https://www.cplusplus.com/reference/fstream/ofstream/

src/utils/io/serial.cpp
```diff
@@ -0,0 +1,138 @@
...
+  }
+
+  uint8_t checksum = payloadSize;
+
+  // Write the HEADER
+  sendData(HEADER);
+  sendData(payloadSize);
+
+  // Write the payload
+  for (int i = 0; i < payloadSize; i++) {
```
maxguy2001, line 49:
```cpp
for (size_t i = 0; i < payloadSize; ++i)
```

src/utils/io/serial.cpp
```diff
@@ -0,0 +1,138 @@
...
+    DataLength   (1 byte)            - The size of the payload
+    Data         (DataLength bytes)  - Actual payload
+    Checksum     (1 byte)            - Message integrity check
+
+    Total message size = (payload size + 3) bytes;
+*/
+
+namespace hyped::utils::io {
+
+SerialProtocol::SerialProtocol(int serial, uint8_t *payload, uint8_t payloadSize)
```
maxguy2001, line 22: `const`

src/utils/io/serial.cpp
```diff
@@ -0,0 +1,138 @@
...
+{
+  setSerial(serial);
+  this->payload     = payload;
+  this->payloadSize = payloadSize;
+  inputBuffer       = (uint8_t *)malloc(payloadSize);
+}
+
+void SerialProtocol::setSerial(int serial)
```
maxguy2001, line 30: `const`

src/utils/io/serial.cpp
```diff
@@ -0,0 +1,138 @@
...
+    ++bytesRead;
+  }
+}
+
+bool SerialProtocol::serialAvailable()
+{
+  return serial != -1;
+}
+
+void SerialProtocol::sendData(uint8_t data)
```
maxguy2001, line 125: `const`

maxguy2001 (changes requested):

arduino/main.ino
```diff
@@ -0,0 +1,49 @@
...
+
+    payload.wheelEncoderA = 0;
+    payload.wheelEncoderB = 0;
+    payload.wheelEncoderC = 0;
+    payload.wheelEncoderD = 0;
+}
+
+void loop()
+{
+    if(digitalRead(IRSensor1)){
```
MiltFra, line 34: `if (...) {`

MiltFra: You'd probably want to just run clang-format on this...

arduino/main.ino
```diff
@@ -0,0 +1,49 @@
+#include <SoftwareSerial.h>
+#include <SerialProtocol.h>
+#include <StreamSerialProtocol.h>
+#include <ArduinoSerialProtocol.h>
+
+int IRSensor1 = 2;
```
MiltFra, line 6: If these are constants, you probably actually do want to `#define` them (assuming this is C). If it's C++, `constexpr`, you know the drill.

Also, naming...

arduino/main.ino
```diff
@@ -0,0 +1,49 @@
...
+int IRSensor2 = 3;
+int IRSensor3 = 4;
+int IRSensor4 = 5;
+
+// The payload that will be sent to the other device
+struct Payload {
+    int16_t wheelEncoderA;
+    int16_t wheelEncoderB;
```
MiltFra, line 14: `wheel_encoder_a`

MiltFra (changes requested):

src/utils/io/serial.hpp
```diff
@@ -0,0 +1,58 @@
...
+  kWaitingForData = 3,
+  // The size of the received payload doesn't match the expected size
+  kInvalidSize = 4,
+  // The object was received but it is not the same as one sent
+  kInvalidChecksum = 5
+};
+
+enum class BaudRate {
+  kB_300,
+  kB_600,
```
MiltFra, line 28: Strictly speaking, this is not camel case. Just remove the underscore, I reckon.

src/utils/io/serial.hpp
```diff
@@ -0,0 +1,58 @@
...
+  kB_2400,
+  kB_4800,
+  kB_9600,
+  kB_14400,
+  kB_19200,
+  kB_28800
+};
+class SerialProtocol {
+ public:
+  SerialProtocol(std::string serial, BaudRate baud_rate, Logger &log);
```
MiltFra, line 39: `SerialProtocol(const std::string &device, const BaudRate baud_rate)`

src/utils/io/serial.cpp
```diff
@@ -0,0 +1,118 @@
...
+#include <iostream>
+#include <sstream>
+
+namespace hyped::utils::io {
+
+SerialProtocol::SerialProtocol(const std::string serial, const BaudRate baud_rate,
+                               utils::Logger &log)
+    : serial_device_(serial),
+      baud_rate_(baud_rate),
+      log_(log)
```
MiltFra, line 17: Initialise the logger here using the `"SERIAL"` tag and the log level in the system config.

MiltFra: This has been done in other places, e.g. for the pressure sensor.

src/utils/io/serial.cpp
```diff
@@ -0,0 +1,118 @@
...
+
+void SerialProtocol::readData(std::vector<uint8_t> &data)
+{
+  data.clear();
+  if (!serialAvailable()) {
+    log_.error("Serial device not available for read.");
+    return;
+  }
+
+  size_t n = read(serial_, &readBuffer_[0], readBufferSize_B_);
```
MiltFra, line 93: `n` is a bad variable name. Also, `const`.

src/utils/io/serial.cpp
```diff
@@ -0,0 +1,118 @@
...
+
+  if (n < 0) {
+    log_.error("Error reading from serial device.");
+    return;
+  }
+
+  copy(readBuffer_.begin(), readBuffer_.begin() + n, back_inserter(data));
```
MiltFra, line 100: `std::copy` for clarity.

src/utils/io/serial.cpp
MiltFra: Also, this seems like a better way of doing it: https://stackoverflow.com/a/15138839

Note that this doesn't even require a buffer size if we are reading until the EOF (which it seems like is the case).

MiltFra (changes requested):

src/utils/io/serial.cpp
```diff
@@ -8,13 +8,14 @@
 #include <iostream>
 #include <sstream>
 
+#include <utils/system.hpp>
+
 namespace hyped::utils::io {
 
-SerialProtocol::SerialProtocol(const std::string serial, const BaudRate baud_rate,
-                               utils::Logger &log)
+SerialProtocol::SerialProtocol(const std::string serial, const BaudRate baud_rate)
```
MiltFra, line 15: `const std::string &serial`

src/utils/io/serial.cpp
```diff
@@ -90,14 +91,14 @@ void SerialProtocol::readData(std::vector<uint8_t> &data)
     return;
   }
 
-  size_t n = read(serial_, &readBuffer_[0], readBufferSize_B_);
+  const size_t bytesRead = read(serial_, &readBuffer_[0], readBufferSize_B_);
```
MiltFra, line 94: `ssize_t`

MiltFra (changes requested):

src/utils/io/serial.cpp
```diff
@@ -0,0 +1,119 @@
...
+#include <unistd.h>
+
+#include <iostream>
+#include <sstream>
+
+#include <utils/system.hpp>
+
+namespace hyped::utils::io {
+
+SerialProtocol::SerialProtocol(std::string &serial, const BaudRate baud_rate)
```
MiltFra, line 15: `const` (you need it in the header, too)

src/utils/io/serial.cpp
```diff
@@ -0,0 +1,119 @@
...
+};
+
+bool SerialProtocol::serialAvailable()
+{
+  return serial_ > 0;
+}
+
+void SerialProtocol::configureTermios()
+{
+  serial_ = open(serial_device_.c_str(), O_RDWR | O_NOCTTY | O_NDELAY);
```
arjunnaha, line 35: I think some of these lines need comments as the constant names are not obvious (e.g. `O_RDWR`)

src/utils/io/serial.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+  if (serial_ == -1) {
+    log_.error("Unable to open serial device: %s", serial_device_.c_str());
+    return;
+  }
+
+  struct termios options;
+  tcgetattr(serial_, &options);
+  options.c_cflag = CS8 | CLOCAL | CREAD;
```
arjunnaha, line 44: Same here

src/utils/io/serial.hpp
```diff
@@ -0,0 +1,35 @@
...
+#include <stdlib.h>
+#include <string.h>
+#include <termios.h>
+
+#include <vector>
+
+#include <utils/logger.hpp>
+namespace hyped::utils::io {
+
+enum class BaudRate { kB300, kB600, kB1200, kB2400, kB4800, kB9600, kB14400, kB19200, kB28800 };
```
arjunnaha, line 13:
```suggestion
enum class BaudRate { kB300, kB600, kB1200, kB2400, kB4800, kB9600, kB14400, kB19200, kB28800 };

```

arduino/main.ino
arjunnaha, line 6: This seems brittle if we add/remove the number of IR sensors and wheel encoders? Can we not maintain a map of IR sensors and wheel encoders and loop through?

arjunnaha (changes requested):

arduino/main.ino
MiltFra: I disagree. We are only ever going to have four IR sensors. Let's not make it more complicated than it needs to be.

Of course, this raises the question of how we are going to do testing when we require all four to be present.

PR #123, opened 2022-03-09
Dynamic Traction Demonstration State Machine

ishmis:
## Description
Implementing demo_state_machine. Added missing instances of `kPreReady` in main.cpp under Brakes and `fake_trajectory.cpp` under Sensors. 

- [x] Adding gtests
- [x] Acceleration Time
- [x] Braking Command 
- [x] Add UI

### DIFF
--- a/src/demo_state_machine/ui.cpp
+++ b/src/demo_state_machine/ui.cpp
@@ -16,6 +16,7 @@ Ui::Ui()
   addBrakingCommand();
   addHighPowerOffCommand();
   addShutDownCommand();
+  addStopCommand(); 
   addHelpCommand();
   addQuitCommand();
 }
@@ -43,8 +44,9 @@ void Ui::printAvailableCommands()
 void Ui::readAndHandleCommand()
 {
   const auto current_state = data_.getStateMachineData().current_state;
-  if (current_state == data::State::kPreCalibrating || current_state == data::State::kReady
-      || current_state == data::State::kAccelerating || current_state == data::State::kCruising
+  if (   current_state == data::State::kPreCalibrating || current_state == data::State::kReady
+      || current_state == data::State::kAccelerating   || current_state == data::State::kCruising 
+      || current_state == data::State::kNominalBraking 
       || current_state == data::State::kFinished) {
     std::cout << "> ";
     std::string command;
@@ -121,6 +123,18 @@ void Ui::giveHighPowerOffCommand()
 void Ui::addHighPowerOffCommand()
 {
   addCommand({"hpoff", "Allows pod to enter Nominal Braking", [this]() { giveHighPowerOffCommand(); }});
+}
+
+void Ui::giveStopCommand()
+{
+  telemetry_data_                  = data_.getTelemetryData();
+  telemetry_data_.stop_command     = true;
+  data_.setTelemetryData(telemetry_data_);
+}
+
+void Ui::addStopCommand()
+{
+  addCommand({"stop", "Allows pod to enter Finished", [this]() { giveStopCommand(); }});
 }
 
 void Ui::giveShutDownCommand()
--- a/src/demo_state_machine/ui.hpp
+++ b/src/demo_state_machine/ui.hpp
@@ -45,12 +45,14 @@ class Ui {
   void addBrakingCommand();
   void addShutDownCommand();
   void addHighPowerOffCommand();
+  void addStopCommand(); 
 
   void giveCalibrationCommand();
   void giveLaunchCommand();
   void giveBrakingCommand();
   void giveHighPowerOffCommand(); 
   void giveShutDownCommand();
+  void giveStopCommand();
 };
 
 }  // namespace hyped::demo_state_machine


SnickeyX:
@MiltFra its not complete yet, just wanted to check if what I'm done so far is okay

MiltFra (approved):
Looking good so far.

adenhaus (approved):

src/demo_state_machine/main.cpp
```diff
@@ -0,0 +1,40 @@
+#include "main.hpp"
+
+#include <cstdint>
```
maxguy2001, line 3: is this needed?

SnickeyX: oh, good shout, don't think this is needed

MiltFra (changes requested):

kshxtij (approved):

SnickeyX:
demo UI does not work atm, still working on it

src/demo_state_machine/UI.cpp
```diff
@@ -0,0 +1,132 @@
+#include "UI.hpp"
+
+#include <fstream>
+#include <iostream>
+#include <sstream>
+
+namespace hyped::demo_state_machine {
+
+UI::UI()
```
MiltFra, line 9: `Ui`

src/demo_state_machine/UI.cpp
```diff
@@ -46,9 +46,9 @@ void UI::printAvailableCommands()
   }
 }
 
-void UI::readAndHandleCommand()
+void Ui::readAndHandleCommand()
 {
-  auto current_state = data_.getStateMachineData().current_state;
+  auto const current_state = data_.getStateMachineData().current_state;
```
MiltFra, line 51: while I think the `T const var_name` syntax has its merits, especially with pointers and references, we have somewhat agreed to put the `const` first so we shall continue to do that. consistency is more important than optimising every line.

some examples on why i believe putting the type first is better (even though i don't want you to do it)

```cpp
int const &x; // const & as a const-reference modifier
int const *const x; // one type but more than one const; 
                    // also it's clear that, reading from right to left,
                    // the pointer is constant as well as the underlying memory
```

MiltFra:
@kshxtij please ping me once you have thoroughly reviewed this.

src/demo_state_machine/UI.cpp
```diff
@@ -0,0 +1,138 @@
+#include "UI.hpp"
```
MiltFra, line 1: file names are lowercase

src/demo_state_machine/state.cpp
```diff
@@ -0,0 +1,315 @@
+#include "state.hpp"
+
+#include <chrono>
+
+namespace hyped {
```
maxguy2001, line 5: `namespace hyped::demo_state_machine` ?

src/demo_state_machine/state.cpp
```diff
@@ -0,0 +1,315 @@
...
+data::State Accelerating::enum_value_       = data::State::kAccelerating;
+char Accelerating::string_representation_[] = "Accelerating";
+
+State *Accelerating::checkTransition(Logger &log)
+{
+  updateModuleData();
+
+  const uint64_t time_now    = utils::Timer::getTimeMicros();
+  const int64_t time_elapsed = time_now - instance_.acceleration_start_;
+
```
maxguy2001, line 154: If these are both timestamps, why are they different types?

src/demo_state_machine/transitions.cpp
```diff
@@ -0,0 +1,135 @@
...
+  if (!telemetry_data.emergency_stop_command) return false;
+
+  return true;
+}
+
+//--------------------------------------------------------------------------------------
+// Navigation Data Events
+//--------------------------------------------------------------------------------------
+
+bool checkEnteredBrakingZone(utils::Logger &log, const data::Navigation &nav_data)
```
maxguy2001, line 109: Do we need this in the demo state machine?

src/demo_state_machine/transitions.cpp
```diff
@@ -0,0 +1,135 @@
...
+{
+  if (nav_data.velocity < Navigation::kMaximumVelocity) return false;
+
+  log.info("reached maximum velocity");
+  return true;
+}
+
+bool checkPodStopped(utils::Logger &log, const data::Navigation &nav_data)
+{
+  if (nav_data.velocity > 0) return false;
```
maxguy2001, line 129: I'm not 100% sure it will be exactly zero since it is type float. Maybe we could use a small number as a tolerance?

src/demo_state_machine/transitions.hpp
```diff
@@ -0,0 +1,106 @@
+#pragma once
+
+#include "state.hpp"
+
+#include <data/data.hpp>
+#include <utils/logger.hpp>
+
+namespace hyped {
+
+namespace demo_state_machine {
```
maxguy2001, line 10: `namespace hyped::demo_state_machine`?

src/demo_state_machine/transitions.cpp
```diff
@@ -0,0 +1,135 @@
...
+{
+  data::nav_t remaining_distance = Navigation::kRunLength - nav_data.displacement;
+  data::nav_t required_distance  = nav_data.braking_distance + Navigation::kBrakingBuffer;
+  if (remaining_distance > required_distance) return false;
+
+  log.info("entered braking zone");
+  return true;
+}
+
+bool checkReachedMaxVelocity(utils::Logger &log, const data::Navigation &nav_data)
```
maxguy2001, line 119: Do we need this? I thought the pod wasn't really moving in demo?

src/utils/io/gpio.cpp
```diff
@@ -123,7 +123,7 @@ void Gpio::uninitialise()
 void Gpio::exportGPIO()
 {
   if (!initialised_) {
-    log_.error("servise has not been initialised");
+    log_.error("service has not been initialised");
```
maxguy2001, line 126: 😍

maxguy2001 (changes requested):

src/demo_state_machine/transitions.cpp
SnickeyX: nope, will remove

src/demo_state_machine/transitions.hpp
SnickeyX: syntactic difference, will change it still

src/demo_state_machine/state.cpp
SnickeyX: `getTimeMacros()` gives a uint64_t so I have to use it, but for `time_elapsed` we need to worry about it being signed (positive specifically) coz negative time difference should not happen and we should know that's the case

src/demo_state_machine/transitions.cpp
SnickeyX: nope, will remove

PR #131, opened 2022-03-11
MTC - Static rpm regulator

TomLonergan03:
## Description
Converted rpm regulator to be static


TomLonergan03:
This is being weird. I get errors with eigen somethings on my pc when I build but it works fine here.

adenhaus (approved):

maxguy2001 (approved):

PR #133, opened 2022-03-21
Add VSCode debugging task

arjunnaha:
## Description
Adds the debugging task that integrates with the build system:
image

Please can reviewers check this works on their local machine

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -1,0 +1,53 @@
+{
+  "tasks": [
+    {
+      "label": "cmake build",
+      "command": "cmake",
+      "args": [
+        "-S .",
+        "-B gui-build",
+        "-DPEDANTIC=ON",
+        "-DRELEASE=ON",
+        "-DCOVERAGE=ON",
+        "-DCROSS=OFF",
+        "-DCMAKE_CXX_COMPILER=/usr/bin/clang++",
+        "-DCMAKE_BUILD_TYPE=Debug",
+      ],
+      "options": {
+        "cwd": "${workspaceFolder}",
+      },
+      "problemMatcher": [
+        "$gcc"
+      ],
+    },
+    {
+      "label": "make build",
+      "command": "make",
+      "args": [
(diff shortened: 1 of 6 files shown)


.vscode/tasks.json
```diff
@@ -0,0 +1,53 @@
...
+  "tasks": [
+    {
+      "label": "cmake build",
+      "command": "cmake",
+      "args": [
+        "-S .",
+        "-B gui-build",
+        "-DPEDANTIC=ON",
+        "-DRELEASE=ON",
+        "-DCOVERAGE=ON",
```
MiltFra, line 11: Do we need this to be on?

.vscode/tasks.json
```diff
@@ -0,0 +1,53 @@
+{
+  "tasks": [
+    {
+      "label": "cmake build",
+      "command": "cmake",
+      "args": [
+        "-S .",
+        "-B gui-build",
+        "-DPEDANTIC=ON",
+        "-DRELEASE=ON",
```
MiltFra, line 10: This should definitely be off.

.vscode/tasks.json
```diff
@@ -0,0 +1,53 @@
+{
+  "tasks": [
+    {
+      "label": "cmake build",
+      "command": "cmake",
+      "args": [
+        "-S .",
+        "-B gui-build",
```
MiltFra, line 8: Not sure this is the right folder? Don't we want to keep this reserved for the GUI? I'd do `./build` instead.

configurations/fake_config.json
```diff
@@ -1,44 +1,4 @@
 {
-  "system": {
```
MiltFra, line 2: Why is this being deleted here?

MiltFra (changes requested):

.vscode/tasks.json
arjunnaha: Fixed :)

MiltFra (approved):

PR #137, opened 2022-03-30
SNS: I2C Interface

kshxtij:
## Description

Implement I2C interface for Pressure Sensor.

### DIFF
--- a/src/utils/io/i2c.cpp
+++ b/src/utils/io/i2c.cpp
@@ -16,12 +16,12 @@
 
 namespace hyped::utils::io {
 
-I2c::I2c(const uint8_t busAddr)
+I2c::I2c(const uint8_t bus_address)
     : log_(utils::Logger("I2C", utils::System::getSystem().config_.log_level)),
-      sensor_addr_(0)
+      sensor_address_(0)
 {
   char path[13];  // up to "/dev/i2c-255"
-  sprintf(path, "/dev/i2c-%d", busAddr);
+  sprintf(path, "/dev/i2c-%d", bus_address);
   fd_ = open(path, O_RDWR, 0);
   if (fd_ < 0) { log_.error("Could not open i2c device"); };
 }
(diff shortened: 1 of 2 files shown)


src/utils/io/i2c.cpp
```diff
@@ -0,0 +1,78 @@
...
+}
+
+void I2C::readData(const uint32_t addr, uint8_t *data, const uint8_t len)
+{
+  if (fd_ < 0) {
+    log_.error("Could not find i2c device");
+    return;
+  }
+
+  if (sensor_addr_ != addr) setSensorAddress(addr);
```
MiltFra, line 55: `{ /* ... */ }`

src/utils/io/i2c.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+  sensor_addr_ = addr;
+  int ret      = ioctl(fd_, I2C_SLAVE, addr);
+  if (ret < 0) {
+    log_.error("Could not set sensor address");
+    return;
+  }
+}
+
+void I2C::readData(const uint32_t addr, uint8_t *data, const uint8_t len)
```
MiltFra, line 48: This function currently has no way of reporting an error. Since we're going with the standard Unix style here, good practice would be to have it return an integer. Then negative means failure and you also get a confirmation of how many bytes were read.

src/utils/io/i2c.cpp
```diff
@@ -0,0 +1,78 @@
...
+{
+  if (fd_ < 0) {
+    log_.error("Could not find i2c device");
+    return;
+  }
+
+  if (sensor_addr_ != addr) setSensorAddress(addr);
+
+  int ret = read(fd_, data, len);
```
MiltFra, line 57: the type of `read` is 

```cpp
ssize_t read(int fildes, void *buf, size_t nbyte)
```

the file descriptor and the data are fine, but `len` should be `size_t`. it's in line with what we're doing elsewhere, too.

also, ret should be `ssize_t`. to make your life easy, why not just use `const auto`?

src/utils/io/i2c.cpp
```diff
@@ -0,0 +1,78 @@
...
+  }
+}
+
+void I2C::writeData(const uint32_t addr, uint8_t *data, const uint8_t len)
+{
+  if (fd_ < 0) {
+    log_.error("Could not find i2c device");
+    return;
+  }
+  if (sensor_addr_ != addr) setSensorAddress(addr);
```
MiltFra, line 70: see above

src/utils/io/i2c.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+  int ret = write(fd_, data, len);
```
MiltFra, line 72: see above

src/utils/io/i2c.hpp
```diff
@@ -0,0 +1,24 @@
...
+#include <utils/logger.hpp>
+
+namespace hyped::utils::io {
+
+class I2C {
+ public:
+  I2C();
+  ~I2C();
+
+  void readData(uint32_t addr, uint8_t *data, uint8_t len);
```
MiltFra, line 12: `const` where possible

src/utils/io/i2c.hpp
```diff
@@ -0,0 +1,24 @@
+#include <cstdint>
+
+#include <utils/logger.hpp>
+
+namespace hyped::utils::io {
+
+class I2C {
```
MiltFra, line 7: There is a philosophical question about the casing of digits here, but I read I2C as a single word so the class should be called `I2c`. This is being changed for ADC and GPIO and SPI in the debug binary PR.

src/utils/io/i2c.hpp
```diff
@@ -0,0 +1,24 @@
...
+ public:
+  I2C();
```
MiltFra, line 9: Do we only have one I2C port? Otherwise, we may want to make it possible to choose one.

src/utils/io/i2c.cpp
```diff
@@ -0,0 +1,78 @@
...
+#endif
+
+#include "utils/logger.hpp"
+#include "utils/system.hpp"
+
+namespace hyped::utils::io {
+
+I2C::I2C()
+    : log_(utils::Logger("I2C", utils::System::getSystem().config_.log_level)),
+      fd_(0),
```
MiltFra, line 21: Given that you check leq 0 below, this should be -1. However, you don't need to initialise it at all as you're doing that in the body of the constructor.

src/utils/io/i2c.cpp
MiltFra: also, do we expect the same I2C instance to be used for different sensor addresses? otherwise, we could just restrict that and do the `ioctl` call on init?

MiltFra (changes requested):
Overall quite good. Bit of style and a bit of usability, but that should be over quickly. Good job!

src/utils/io/i2c.cpp
kshxtij: Yes because the call to get temperature and pressure would be 2 different sensor addressees even tho they are connected on the same bus. I know the wording is ambiguous but i followed the same naming convention as most online documentation about i2c.

MiltFra: Would it then be a good idea to always call ioctl again? Or is that an expensive operation?

kshxtij: Multiple i2c instances on the same bus would fail as you wouldn't be allowed to open the same file more than once while the process is running iirc. Also since we are passing in the bus manually, this would also be a common sense thing? as each bus can physically be only connected one sensor.

MiltFra (changes requested):

src/utils/io/i2c.cpp
```diff
@@ -0,0 +1,81 @@
...
+  fd_ = open(path, O_RDWR, 0);
+  if (fd_ < 0) { log_.error("Could not open i2c device"); };
+}
+
+I2c::~I2c()
+{
+  close(fd_);
+}
+
+void I2c::setSensorAddress(uint32_t address)
```
maxguy2001, line 34: `const`?

SnickeyX (approved):

MiltFra (approved):

PR #141, opened 2022-04-01
DEBUG: Debug binary #2

mifrandir:
- [x] Direct CAN
  - [x] sending
  - [x] receiving

### DIFF
--- a/src/debugging/can_listener.cpp
+++ b/src/debugging/can_listener.cpp
@@ -1,0 +1,41 @@
+#include "can_listener.hpp"
+
+#include <utils/system.hpp>
+
+namespace hyped::debugging {
+
+CanListener::CanListener() : log_("CAN-LISTENER", utils::System::getSystem().config_.log_level)
+{
+}
+
+void CanListener::processNewData(utils::io::can::Frame &message)
+{
+  log_.info("Received CAN frame: {node_id=%u, extended=%s, len=%u, data=%x %x %x %x %x %x %x %x}",
+            message.id, message.id, message.len, message.data[0], message.data[1], message.data[2],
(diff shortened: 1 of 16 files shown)


src/debugging/can_listener.cpp
```diff
@@ -0,0 +1,41 @@
...
+}
+
+void CanListener::processNewData(utils::io::can::Frame &message)
+{
+  log_.info("Received CAN frame: {node_id=%u, extended=%s, len=%u, data=%x %x %x %x %x %x %x %x}",
+            message.id, message.id, message.len, message.data[0], message.data[1], message.data[2],
+            message.data[3], message.data[4], message.data[5], message.data[6], message.data[7]);
+}
+
+bool CanListener::hasId(uint32_t id, bool extended)
```
kshxtij, line 18: const?

src/debugging/can_listener.cpp
```diff
@@ -0,0 +1,41 @@
...
+
+#include <utils/system.hpp>
+
+namespace hyped::debugging {
+
+CanListener::CanListener() : log_("CAN-LISTENER", utils::System::getSystem().config_.log_level)
+{
+}
+
+void CanListener::processNewData(utils::io::can::Frame &message)
```
kshxtij, line 11: could this be const?

src/debugging/can_listener.cpp
MiltFra: no, because of the CAN implementation... i considered modifying it but I'm not willing to go down that rabbit hole just yet

src/debugging/can_listener.cpp
MiltFra: see above

kshxtij (approved):

PR #142, opened 2022-04-07
Nav: Wheel encoder outlier detection

maxguy2001:
Outlier detection system for the wheel encoders. Based on the current system used for IMUs.

### DIFF
--- a/src/navigation/navigation.cpp
+++ b/src/navigation/navigation.cpp
@@ -198,7 +198,7 @@ void Navigation::calibrateGravity()
 
 void Navigation::queryWheelEncoders()
 {
-  std::array<hyped::data::nav_t, data::Sensors::kNumEncoders> encoder_data_array;
+  EncoderArray encoder_data_array;
   const auto encoder_data = data_.getSensorsWheelEncoderData();
 
   uint32_t sum                   = 0;
--- a/src/navigation/navigation.cpp
+++ b/src/navigation/navigation.cpp
@@ -198,8 +198,8 @@ void Navigation::calibrateGravity()
 
 void Navigation::queryWheelEncoders()
 {
-  const auto encoder_data = data_.getSensorsWheelEncoderData();
   NavigationArray encoder_data_array;
+  const auto encoder_data = data_.getSensorsWheelEncoderData();
 
   data::nav_t sum                = 0;
   data::nav_t encoder_data_value = 0;
@@ -439,11 +439,12 @@ void Navigation::wheelEncoderOutlierDetection(NavigationArray &data_array,
   for (std::size_t i = 0; i < data::Sensors::kNumEncoders; ++i) {
     const auto exceeds_limits = data_array.at(i) < lower_limit || data_array.at(i) > upper_limit;
     if (exceeds_limits && is_encoder_reliable_.at(i)) {
-      log_.debug("Outlier detected in Encoder %d, reading: %.3f not in [%.3f, %.3f]. Updated to %.3f",
-                 i + 1, data_array.at(i), lower_limit, upper_limit, quartile_bounds.at(1));
+      log_.debug(
+        "Outlier detected in Encoder %d, reading: %.3f not in [%.3f, %.3f]. Updated to %.3f", i + 1,
+        data_array.at(i), lower_limit, upper_limit, quartile_bounds.at(1));
       data_array.at(i) = quartile_bounds.at(1);
       encoder_outlier_counter_.at(i)++;
-      // If this counter exceeds some threshold then that IMU is deemed unreliable
+      // If this counter exceeds some threshold then that encoder is deemed unreliable
       if (encoder_outlier_counter_.at(i) > 1000 && is_encoder_reliable_.at(i)) {
         is_encoder_reliable_.at(i) = false;
         ++num_outlier_encoders_;
--- a/src/navigation/navigation.cpp
+++ b/src/navigation/navigation.cpp
@@ -199,14 +199,20 @@ void Navigation::calibrateGravity()
 void Navigation::queryWheelEncoders()
 {
   const auto encoder_data = data_.getSensorsWheelEncoderData();
+  NavigationArray encoder_data_array;
 
-  data::nav_t sum = 0;
+  data::nav_t sum                = 0;
+  data::nav_t encoder_data_value = 0;
+
   for (size_t i = 0; i < encoder_data.size(); ++i) {
-    sum += static_cast<data::nav_t>(encoder_data.at(i).value);
+    encoder_data_value = static_cast<data::nav_t>(encoder_data.at(i).value);
+    sum += encoder_data_value;
+    encoder_data_array.at(i) = encoder_data_value;
   }
 
   const data::nav_t average   = sum / encoder_data.size();
   encoder_displacement_.value = average * data::Navigation::kWheelCircumfrence;
+  wheelEncoderOutlierDetection(encoder_data_array, kInterQuartileScaler);
 }
 
 void Navigation::queryImus()
@@ -338,8 +344,8 @@ void Navigation::logWrite()
   write_to_file_ = true;
 }
 
-Navigation::QuartileBounds Navigation::calculateQuartiles(NavigationArray &data_array,
-                                                          bool imu_quartiles)
+Navigation::QuartileBounds Navigation::calculateQuartiles(const auto data_array,
+                                                          const bool imu_quartiles)
 {
   std::vector<data::nav_t> data_vector;
   std::array<data::nav_t, 3> quartile_bounds;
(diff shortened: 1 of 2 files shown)


adenhaus (approved):

src/navigation/navigation.cpp
```diff
@@ -334,35 +344,58 @@ void Navigation::logWrite()
...
-  if (num_outlier_imus_ == 0) {
-    quartile_bounds.at(1) = (data_vector.at(1) + data_vector.at(2)) / 2.;
-  } else if (num_outlier_imus_ == 1) {
-    quartile_bounds.at(1) = data_vector.at(1);
-  } else {
-    auto navigation_data          = data_.getNavigationData();
-    navigation_data.module_status = data::ModuleStatus::kCriticalFailure;
-    data_.setNavigationData(navigation_data);
-    log_.error("At least two IMUs no longer reliable, entering CriticalFailure.");
+  if (imu_quartiles == true) {
```
SnickeyX, line 353: you could just do `if(imu_quartiles)` here and if that's not intuitive then maybe change the parameter name (e.g `has_imu_quartiles`)

SnickeyX (changes requested):

src/navigation/navigation.cpp
```diff
@@ -334,35 +344,58 @@ void Navigation::logWrite()
...
+      quartile_bounds.at(1) = (data_vector.at(1) + data_vector.at(2)) / 2.;
+    } else if (num_outlier_imus_ == 1) {
+      quartile_bounds.at(1) = data_vector.at(1);
+    } else {
+      auto navigation_data          = data_.getNavigationData();
+      navigation_data.module_status = data::ModuleStatus::kCriticalFailure;
+      data_.setNavigationData(navigation_data);
+      log_.error("At least two IMUs no longer reliable, entering CriticalFailure.");
+    }
+  } else if (imu_quartiles == false) {
```
SnickeyX, line 372: an `else` should work here just fine since there are no other else if cases in the outer-most if-statement

src/navigation/navigation.cpp
```diff
@@ -392,6 +425,40 @@ void Navigation::imuOutlierDetection(NavigationArray &data_array, const data::na
...
+  const std::array<data::nav_t, 3> quartile_bounds = calculateQuartiles(data_array, false);
+
+  // find the thresholds
+  // clip IQR to upper bound to avoid issues with very large outliers
+  const auto iqr = std::min(quartile_bounds.at(2) - quartile_bounds.at(0), kMaxInterQuartileRange);
+  const auto upper_limit = quartile_bounds.at(2) + threshold * iqr;
+  const auto lower_limit = quartile_bounds.at(0) - threshold * iqr;
+  // replace any outliers with the median
+  for (std::size_t i = 0; i < data::Sensors::kNumEncoders; ++i) {
+    const auto exceeds_limits = data_array.at(i) < lower_limit || data_array.at(i) > upper_limit;
```
SnickeyX, line 440: this could just be `const bool exceeds_limits`

SnickeyX (changes requested):

src/navigation/navigation.cpp
```diff
@@ -392,6 +425,40 @@ void Navigation::imuOutlierDetection(NavigationArray &data_array, const data::na
   }
 }
 
+void Navigation::wheelEncoderOutlierDetection(NavigationArray &data_array,
+                                              const data::nav_t threshold)
```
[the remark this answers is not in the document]
MiltFra: Lastly, why are we even doing this thing where we modify this data array? Would a more functional style not be better? I.e. take in "raw data", do calculations and raise errors as required, and then return an adjusted array. Would be more readable I believe.

MiltFra (changes requested):

src/navigation/navigation.cpp
```diff
@@ -334,35 +344,58 @@ void Navigation::logWrite()
   write_to_file_ = true;
 }
 
-Navigation::QuartileBounds Navigation::calculateImuQuartiles(NavigationArray &data_array)
+Navigation::QuartileBounds Navigation::calculateQuartiles(const auto data_array,
+                                                          const bool get_imu_quartiles)
```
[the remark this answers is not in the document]
maxguy2001: yeah I wasn't sure about sticking it all in 1 function but it just felt more contained, will switch to 2 functions

arjunnaha (approved):

src/navigation/navigation.cpp
```diff
@@ -194,15 +198,21 @@ void Navigation::calibrateGravity()
 
 void Navigation::queryWheelEncoders()
 {
+  EncoderArray encoder_data_array;
   const auto encoder_data = data_.getSensorsWheelEncoderData();
 
-  data::nav_t sum = 0;
+  uint32_t sum                   = 0;
+  data::nav_t encoder_data_value = 0;
```
MiltFra, line 205: This is unused (surprised clang+Werror didn't catch this?). Also, I would restructure this as follows:
```cpp
const auto encoder_data = data_.getSensorsWheelEncoderData();

EncoderArray encoder_data_array;
uint32_t sum = 0;
for (size_t i = 0; i < encoder_data.size(); ++i) {
    sum += encoder_data.at(i).value;
    encoder_data_array.at(i) = encoder_data.at(i).value;
}
```
Alternatively,  if you're feeling fancy
```cpp
const auto encoder_data = data_.getSensorsWheelEncoderData();

EncoderArray encoder_data_array;
std::transform(encoder_data.begin(), encoder_data.end(), encoder_data_array.begin(),
               [](const auto &data) { return data.value; });
const auto sum = std::accumulate(encoder_data_array.begin(), encoder_data_array.end(), 0);
```

src/navigation/navigation.hpp
```diff
@@ -27,6 +27,7 @@ class Navigation {
   using NavigationArrayOneFaulty = std::array<data::nav_t, data::Sensors::kNumImus - 1>;
   using FilterArray              = std::array<KalmanFilter, data::Sensors::kNumImus>;
   using QuartileBounds           = std::array<data::nav_t, 3>;
+  using EncoderArray             = std::array<data::nav_t, data::Sensors::kNumEncoders>;
```
MiltFra, line 30: shouldn't this be `uint32_t`?

src/navigation/navigation.cpp
```diff
@@ -360,9 +370,36 @@ Navigation::QuartileBounds Navigation::calculateImuQuartiles(NavigationArray &da
   return quartile_bounds;
 }
 
-void Navigation::imuOutlierDetection(NavigationArray &data_array, const data::nav_t threshold)
+Navigation::QuartileBounds Navigation::calculateEncoderQuartiles(const EncoderArray &data_array)
 {
-  std::array<data::nav_t, 3> quartile_bounds = calculateImuQuartiles(data_array);
+  std::vector<data::nav_t> data_vector;
+  std::array<data::nav_t, 3> quartile_bounds;
```
MiltFra, line 376: Move this further down towards its first use.

src/navigation/navigation.cpp
```diff
@@ -360,9 +370,36 @@ Navigation::QuartileBounds Navigation::calculateImuQuartiles(NavigationArray &da
   return quartile_bounds;
 }
 
-void Navigation::imuOutlierDetection(NavigationArray &data_array, const data::nav_t threshold)
+Navigation::QuartileBounds Navigation::calculateEncoderQuartiles(const EncoderArray &data_array)
 {
-  std::array<data::nav_t, 3> quartile_bounds = calculateImuQuartiles(data_array);
+  std::vector<data::nav_t> data_vector;
```
MiltFra, line 375: Again, this should be `uint32_t`, right?

src/navigation/navigation.cpp
```diff
@@ -360,9 +370,36 @@ Navigation::QuartileBounds Navigation::calculateImuQuartiles(NavigationArray &da
...
+    = (data_vector.at(data_vector.size() - 2) + data_vector.at(data_vector.size() - 1)) / 2.;
+  if (num_outlier_encoders_ == 0) {
+    quartile_bounds.at(1) = (data_vector.at(1) + data_vector.at(2)) / 2.;
+  } else if (num_outlier_encoders_ == 1) {
+    quartile_bounds.at(1) = data_vector.at(1);
+  } else {
+    auto navigation_data          = data_.getNavigationData();
+    navigation_data.module_status = data::ModuleStatus::kCriticalFailure;
+    data_.setNavigationData(navigation_data);
+    log_.error("At least two Encoders no longer reliable, entering CriticalFailure.");
```
MiltFra, line 394: There clearly is an error here but we still return something without indicating to the caller that this is invalid. The return type of this function should be `std::optional<Navigation::QuartileBounds`; then you can return `std::nullopt` in the failure case.

MiltFra (changes requested):

src/navigation/navigation.cpp
maxguy2001: I was not feeling fancy :(

maxguy2001: I was not feeling fancy :(

src/navigation/navigation.cpp
maxguy2001: the same issue exists in the equivalent imu function by the looks of it but we've written the fail state directly to data so once this iteration of the `nav.navigate()` loop finishes, the main function will switch to failure state and the breaks will be activated so this function will basically be ignored while we brake?

maxguy2001: the same issue exists in the equivalent imu function by the looks of it but we've written the fail state directly to data so once this iteration of the `nav.navigate()` loop finishes, the main function will switch to failure state and the breaks will be activated so this function will basically be ignored while we brake?

MiltFra (approved):

PR #143, opened 2022-04-08
Sns susptemp

supermegho:
Added Brakes and Suspension temperature sensor code. 

Additional calculation added to temperature sensor code based on analogue calculation.

Shut down pod sequence added to main.cpp if we go over 85 degrees Celcius.

### DIFF
--- a/src/sensors/main.cpp
+++ b/src/sensors/main.cpp
@@ -132,6 +132,19 @@ void Main::checkTemperature()
   }
 }
 
+void Main::checkSuspensionTemperature()
+{
+  Temperature_susp_->run();  // not a thread
+
+  const auto temperature_susp = temperature_susp_->getData();  // *C
+  if (temperature_susp > 85) {
+    log_.info("Suspension Temperature (%u) exceeds maximum value (%d)", temperature_susp, 85);
+    auto sensors_data          = data_.getSensorsData();
+    sensors_data.module_status = data::ModuleStatus::kCriticalFailure;
+    data_.setSensorsData(sensors_data);
+  }
+}
+
 void Main::checkAmbientPressure()
 {
   ambient_pressure_->run();  // not a thread
--- a/src/sensors/temperature_amb.cpp
+++ b/src/sensors/temperature_amb.cpp
@@ -1,0 +1,36 @@
+#include "temperature.hpp"
+
+#include <stdio.h>
+
+#include <utils/system.hpp>
+
+namespace hyped::sensors {
+
+Temperature::Temperature(const uint8_t pin)
+    : pin_(pin),
+      log_("TEMPERATURE", utils::System::getSystem().config_.log_level_sensors)
+{
+}
+
+void Temperature::run()
+{
+  uint16_t raw_value = pin_.read();
+  log_.debug("raw value: %d", raw_value);
+  temperature_data_.temperature = scaleData(raw_value);
+  log_.debug("scaled value: %d", temperature_data_.temperature);
+  temperature_data_.operational = true;
+}
+
+int8_t Temperature::scaleData(const uint8_t raw_value)
+{
+  // convert to degrees C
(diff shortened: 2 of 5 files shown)


src/sensors/main.cpp
```diff
@@ -132,6 +132,19 @@ void Main::checkTemperature()
   }
 }
 
+void Main::checkSuspensionTemperature()
```
kshxtij, line 135: You've defined this but you aren't calling it anywhere yet

src/sensors/temperature_susp.hpp
```diff
@@ -0,0 +1,55 @@
...
+#include "interface.hpp"
+
+#include <cstdint>
+
+#include <data/data.hpp>
+#include <utils/io/adc.hpp>
+#include <utils/logger.hpp>
+namespace hyped::sensors {
+
+class Temperature_susp : public ITemperature_susp {
```
kshxtij, line 12: ITemperature_susp isn't defined anywhere +  define the interface within the header instead of interface.hpp

src/sensors/temperature_susp.cpp
```diff
@@ -0,0 +1,36 @@
...
+
+#include <stdio.h>
+
+#include <utils/system.hpp>
+
+namespace hyped::sensors {
+
+Temperature_susp::Temperature_susp(const uint8_t pin)
+    : pin_(pin),
+      log_("TEMPERATURE", utils::System::getSystem().config_.log_level_sensors)
```
kshxtij, line 11: "Brakes and Suspension Temperature"

src/sensors/temperature_amb.cpp
```diff
@@ -0,0 +1,36 @@
...
+
+#include <stdio.h>
+
+#include <utils/system.hpp>
+
+namespace hyped::sensors {
+
+Temperature::Temperature(const uint8_t pin)
+    : pin_(pin),
+      log_("TEMPERATURE", utils::System::getSystem().config_.log_level_sensors)
```
kshxtij, line 11: "Ambient Temperature"

src/sensors/temperature_susp.hpp
```diff
@@ -0,0 +1,55 @@
...
+  /**
+   * @brief ADC pin
+   */
+  utils::io::ADC pin_;
+  utils::Logger log_;
+
+  /**
+   * @brief int from data structs
+   */
+  data::Temperature_suspData temperature_susp_data_;
```
kshxtij, line 52: Can probably just keep data::TemperatureData and even then you'd need to define it within data

src/sensors/temperature_susp.hpp
kshxtij: Also same naming comment from the other file applies here

src/sensors/main.cpp
kshxtij, line 135: Also needs to be defined in main.hpp

kshxtij (changes requested):
Also within main.cpp, you need

- a from file function, this can just be copy pasted with some minor changes on any of the other FromFile functions (look at [this](https://github.com/Hyp-ed/hyped-2022/blob/70e592942ff9e5519b3d57d656f35e18d2b5cd23/src/sensors/main.cpp#L191) or [this](https://github.com/Hyp-ed/hyped-2022/blob/70e592942ff9e5519b3d57d656f35e18d2b5cd23/src/sensors/main.cpp#L233))
- instantiate instances of the class from the pins
- FakeBrakeAndSuspensionTemp

src/sensors/temperature_susp.hpp
MiltFra: I think both types of temperature managers could be using the same interface `ITemperature`. Then it would make sense to have files `temperature.cpp` and `temperature.hpp` that contain the interface and both classes, `AmbientTemperature` and `BrakeAndSuspsenionTemperature`.

PR #144, opened 2022-04-09
MTC Nucleo board signal generation

TomLonergan03:
This is for model F746ZG

### To do
Get working
- [x] DAC signal generation
- [x] Sine wave generation
- [x] Wave frequency switching
- [ ] Can receiving

Make good
- [ ] DAC signal generation
- [ ] Sine wave generation
- [ ] Wave frequency switching
- [ ] Can receiving


TomLonergan03:
If you want to run this it's a little cursed rn, you need to make a new project in STM32IDE with the .ioc file then replace Core with the one in this pull request. Will try to find a nicer way to do it, but for now this is the easiest way

mifrandir (changes requested):
Show me that it works and I shall approve it even though I don't like most of the code here.

test/state_machine/transitions.test.cpp
```diff
@@ -149,7 +149,7 @@ TEST_F(TransitionFunctionality, handlesNavEmergency)
     motors_data.module_status             = other;
     telemetry_data.emergency_stop_command = false;
     const bool has_emergency              = state_machine::checkEmergency(
-                   log_, brakes_data, nav_data, batteries_data, telemetry_data, sensors_data, motors_data);
+      log_, brakes_data, nav_data, batteries_data, telemetry_data, sensors_data, motors_data);
```
MiltFra, line 152: what's up with this?

PR #145, opened 2022-04-11
Test: Replace pow 

kshxtij:
- Replaces math.h import with cmath
- Replaces pow calls with std::pow


maxguy2001 (approved):
Great read, 10/10, would recommend

MiltFra (approved):

PR #152, opened 2022-06-10
SNS - Changes for IMUs to work

ishmis:
Slight modifications which gets the IMU-20948 to work with our code on the beagleboneblack. 

Giving read and write permissions when opening gpio export file to prevent consistent export errors.

### DIFF
--- a/src/sensors/imu.cpp
+++ b/src/sensors/imu.cpp
@@ -29,7 +29,7 @@ static constexpr uint8_t kPwrMgmt1 = 0x06;  // userbank 0
 static constexpr uint8_t kPwrMgmt2 = 0x07;  // userbank 0
 
 // Configuration
-static constexpr uint8_t kReadFlag = 0x80;  // unable to find in datasheet
+static constexpr uint8_t kReadFlag = 0x80;  // msb is 1 which signifies a read operation for SPI
 
 // Configuration bits Imu
 // constexpr uint8_t kBitsFs250Dps             = 0x00;
@@ -286,12 +286,13 @@ data::ImuData Imu::getData()
       }
     } else {
       log_.debug("Getting Imu data");
-      uint8_t response[8];
+      uint8_t response[6];
       int16_t bit_data;
       float value;
       std::array<float, 3> acceleration_data;
 
-      readBytes(kAccelXoutH, response, 8);
+      // Reading six bytes - first two give x-acceleration, next two give y-acceleration, last two give z-acceleration (high and low byte pairs)
+      readBytes(kAccelXoutH, response, 6); 
       for (size_t i = 0; i < 3; ++i) {
         bit_data                = ((int16_t)response[i * 2] << 8) | response[i * 2 + 1];
         value                   = static_cast<float>(bit_data);
--- a/src/utils/io/gpio.cpp
+++ b/src/utils/io/gpio.cpp
@@ -133,7 +133,7 @@ void Gpio::exportGPIO()
   // let the kernel know we are using this pin
   int fd;
   uint32_t len;
-  fd = open("/sys/class/gpio/export", O_WRONLY);
+  fd = open("/sys/class/gpio/export", O_RDWR);
   if (fd < 0) {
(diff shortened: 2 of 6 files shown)


src/utils/io/spi.cpp
```diff
@@ -96,7 +96,7 @@ Spi &Spi::getInstance()
 
 Spi::Spi(Logger &log) : spi_fd_(-1), hw_(0), ch_(0), log_(log)
 {
-  const char device[] = "/dev/spidev1.0";  // spidev1.0 for SPI0
+  const char device[] = "/dev/spidev0.0";  // spidev1.0 for SPI0
```
kshxtij, line 99: Change the comment to spidev0.0?

kshxtij (changes requested):

src/sensors/imu.cpp
```diff
@@ -286,12 +286,13 @@ data::ImuData Imu::getData()
...
       log_.debug("Getting Imu data");
-      uint8_t response[8];
+      uint8_t response[6];
       int16_t bit_data;
       float value;
       std::array<float, 3> acceleration_data;
 
-      readBytes(kAccelXoutH, response, 8);
+      // Reading six bytes - first two give x-acceleration, next two give y-acceleration, last two give z-acceleration (high and low byte pairs)
+      readBytes(kAccelXoutH, response, 6);
```
mifrandir, line 295: This makes a lot of sense now. Good job!

src/utils/io/spi.cpp
```diff
@@ -96,7 +96,7 @@ Spi &Spi::getInstance()
 
 Spi::Spi(Logger &log) : spi_fd_(-1), hw_(0), ch_(0), log_(log)
 {
-  const char device[] = "/dev/spidev1.0";  // spidev1.0 for SPI0
+  const char device[] = "/dev/spidev0.0";  // spidev0.0 for SPI0
```
mifrandir, line 99: Why was it different to begin with? Odd...

src/utils/io/spi.cpp
```diff
@@ -105,7 +105,7 @@ Spi::Spi(Logger &log) : spi_fd_(-1), hw_(0), ch_(0), log_(log)
   }
 
   // set clock frequency
-  setClock(Clock::k1MHz);
+  setClock(Clock::k500KHz);
```
mifrandir, line 108: Is this required? Did we test it? Has it to do with the new IMUs?

src/utils/io/spi.cpp
```diff
@@ -149,14 +149,17 @@ bool Spi::initialise()
   ch_ = &hw_->ch0;
 
   log_.info("Mapping successfully created %d", sizeof(SPI_HW));
-  log_.info("revision 0x%x", hw_->revision);
+  //log_.info("revision 0x%x", hw_->revision);
```
mifrandir: Remove.

mifrandir (changes requested):
Small things; good stuff otherwise!

src/utils/io/spi.cpp
ishmis: No idea. Found this out using the BBIO library in Python.

src/utils/io/spi.cpp
ishmis: Yes, yes and yes. Although this just needs to be tested with the sensors PCB but with an individual IMU, this worked.

src/utils/io/gpio.cpp
```diff
@@ -133,7 +133,7 @@ void Gpio::exportGPIO()
   // let the kernel know we are using this pin
   int fd;
   uint32_t len;
-  fd = open("/sys/class/gpio/export", O_WRONLY);
+  fd = open("/sys/class/gpio/export", O_RDWR);
```
mifrandir, line 136: You've touched this line now so we need to do this properly. 

```cpp
const auto fd = open(/* ... */);
```

Similar for `len`.

mifrandir (changes requested):

